PR #3, opened 2018-10-26
SL 154/select

casserni:
adds a select component. Leverages the react-select built in styles option to use colors set in the theme


P0lip:
I like the idea of using passing our own theme object. As Marc stated, we could use themeGet to do hat or just forward our own theme prop i.e. `props.theme.components.select` (not sure if that would work though)

casserni:
My only hesitation to overriding their theme object is its very unclear what, for example neutral30 is being used for? Unless we have more friendly names in our theme that just get mapped over to whatever

P0lip:
Well, their naming convention is indeed a bit confusing and somewhat difficult to reason about.
We can either use some more reasonable names (if possible) or just link their legend.

marbemac:
Agree that it's kind of confusing - but I think that's OK for now? Creating a theme will be a rare task - we can create a quick description of the select theme props when we do the readme.

casserni:
updated to leverage select's build in theme and over ride the theme. I attempted to leave sensible comments about which color points to what component items. Ultimately the way they pair colors is a little wonky so we might still need to go back and add some custom logic.

marbemac:
Should we just override and add new theme props (via the original complex method) the minimal amount to make it look good (arrow, for example)?

casserni:
@P0lip as for colorizations in the dark theme, once everything is build out and functional we will choose a color palette and go back and update. For now wanted to chose something that wasn't straight black to show a more complex styling. Used https://blueprintjs.com/docs/ with its dark mode as inspiration

PR #4, opened 2018-10-26
SL-151/scroller

casserni:
adds scroll container that support custom theming

**NOTE**
If using a mac, you may not see these custom scroll bars and instead see the mac replaced one. To see these custom ones you may need to go to 
system preferences -> general -> Show Scrollbars-> always
This may also cause two sets of scroll bars to appear? Need to confirm


marbemac:
Refactored, check it out - feels like a nicer pattern around exposing the inner functions like scrollToBottom, etc via a passed in innerRef (example in story). Also renamed component, figured we might want to rename `ListScroller` to be `ScrollList` for consistency?

src/theme/components.ts
```diff
@@ -37,6 +37,10 @@ export const components: ISectionTheme = {
       border: '#/colors/transparent',
     },
 
+    scrollbar: {
```
P0lip: Could we add bg to https://github.com/stoplightio/ui-kit/blob/master/src/theme/components.ts#L23#L29?
I know it's consumed in a different component, but still, we could try to add it there? WDYT?
If that's not the case, I would consider renaming it to scrollContainer or something similar just to make it more clear.

src/ScrollBox.tsx
```diff
@@ -0,0 +1,110 @@
...
+            right: 10,
+            cursor: 'pointer',
+          }}
+        />
+      )}
+      renderThumbHorizontal={() => {
+        return <Box radius="full" height="6px" bg="scrollbar.bg" cursor="grab" />;
+      }}
+      renderTrackVertical={() => (
+        <div
```
P0lip: Could we use Box here or do we want to avoid theming?

src/theme/components.ts
casserni: I think that we might want to remove the scrollbars section in the top level of the theme. Allowing a user to change width and radius of the thumb is too much customizability. Really just allow users to change color

src/ScrollBox.tsx
casserni: Avoid theming on the track. Always hiding it makes for a cleaner ui. Would be a quick change to expose it though.

src/theme/components.ts
P0lip: Agreed.

marbemac: Agreed - will leave it to you @casserni.

marbemac: Actually this gets back to the whole separating out "user customizable" portion of theme thing. We do want to maintain consistency with the look and feel of these scrollbars and https://github.com/stoplightio/ui-kit/blob/master/src/ListScroller.tsx#L69. Is there a better way than via theme variables that is not hardcoding duplicate values in both places?

casserni:
Updated both ScrollBox and ScrollList to use our custom rendered toolbars. These both offset and hide the native toolbars so we won't need to worry about differences in browser behavior. Note that we should never use overflow scroll EVER and use ScrollBox in its place

src/__stories__/ScrollList.tsx
```diff
@@ -4,30 +4,35 @@ import { array, number, NumberOptions, withKnobs } from '@storybook/addon-knobs'
...
+const rowRenderer = ({ value, key, style }: IScrollListItemProps) => (
   <Box key={key} as="div" {...style as object} borderBottom="2px solid black">
     {value}
   </Box>
 );
 
-storiesOf('ListScroller', module)
+storiesOf('ScrollList', module)
   .addDecorator(withKnobs)
+  // .addDecorator(storyFn => (
```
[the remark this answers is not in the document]
casserni: Good catch

src/ScrollBox.tsx
```diff
@@ -0,0 +1,192 @@
...
+              left: 2,
+            }}
+          />
+        );
+      }}
+      renderThumbHorizontal={({ style }: any) => {
+        return (
+          <Box
+            radius="full"
+            // @ts-ignore
```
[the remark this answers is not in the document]
casserni: Don't left it in by accident

src/ScrollBox.tsx
```diff
@@ -0,0 +1,192 @@
...
+          />
+        );
+      }}
+    >
+      {children}
+    </Scrollbars>
+  );
+};
+
+const useScrollTo = (elementId?: string) => {
```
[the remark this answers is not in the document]
marbemac: Agreed.

src/ScrollList.tsx
```diff
@@ -0,0 +1,133 @@
...
+      {({ height, width }) => (
+        <Box height={height} width={width} overflow="hidden" className={'box'}>
+          <List
+            className={className}
+            rowHeight={rowHeight}
+            rowCount={list.length}
+            rowRenderer={renderRow as ListRowRenderer}
+            noRowsRenderer={noRowsRenderer}
+            onScroll={(e: IOnScroll) => {
+              setClientHeight(e.clientHeight);
```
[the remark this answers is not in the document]
P0lip:
```js
const [{ clientHeight, scrollHeight, scrollTop }, setScrollEvent] = React.useState({});
```

something like that ^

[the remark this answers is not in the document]
casserni: Much cleaner thanks!

src/ScrollBox.tsx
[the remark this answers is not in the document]
casserni: Moved this to the hooks folder and created some other utils

PR #10, opened 2018-11-12
Sl 147/checkbox toggle

rainum:


rainum:
@casserni sounds good! I like custom checkbox idea. I already implemented themed context menu, looks much better than default one 😄

PR #12, opened 2018-11-13
SL-142/popup

P0lip:
~~Started off with some basic CSS-based positioning for now. If an absolute positioning is needed, I'm happy to implement it.~~

Absolute positioning is already in place.


src/Portal.tsx
```diff
@@ -1,12 +1,14 @@
...
 export interface IPortalProps {
-  children: JSX.Element | string | number;
+  children: any;
   className?: string;
 }
 
-export class Portal extends PureComponent<IPortalProps> {
+export class Portal extends React.PureComponent<IPortalProps> {
   private readonly el?: HTMLDivElement;
   private readonly root = typeof document === 'object' ? document.body : null;
```
P0lip: I am not quite sure if we should attach Portal under document.body. Basically, this leads to the situation when theming is slightly broken, since we cannot rely on rules inheritance and we have to style components on our own (see a related story for a reference)

thoughts? @marbemac @lottamus @casserni

rainum: I'm not sure that rules inheritance is applicable to styled components. Usually it adds classes to every component with a proper set of rules. Theming works by overriding some of the rules for every component.

P0lip: In general you are right, but unfortunately, some rules are (or might be) inherited (depending on global styles).
For instance, in this particular case, the color is inherited.
image
This is most certainly caused by the fact Box doesn't set color on its own node, but still - a developer might want to have some global styles for everything under given node.

marbemac: Ah, interesting! There are tradeoffs here, but for simplicity and for now I think the portal just needs to be within the theme provider (and thus have access to the theme). IE if you create a popup with a `<Button bg="error">Foo</Button>` in it, that button should have a red background when rendered in the portal. Does that make sense?

P0lip: Well, I added ThemeConsumer to the Portal, so its child components do have access to the theme. I think it's a nice tradeoff - you can still use theming, but also opt-out if you need to for some reason.
You can see that in action in the story book :)

P0lip: My apologies for dropping a ball here.
I adjusted the stories to show that the rendered content can be actually themed in the way @marbemac suggested.

I also made the final API a bit more end-user friendly, as I wrap the trigger component with required event handlers.

marbemac:
It would be a little work, but could all the state / lifecycle management be made much simpler by switching from es6 class to hooks :)?

marbemac:
Looks good to me! Only suggestion I'd have is that perhaps some of those hooks (like resize) could be extracted out for re-use. Something like https://gist.github.com/gaearon/cb5add26336003ed8c0004c4ba820eae.

@casserni how are you re your review?

src/__tests__/useWindowResize.spec.tsx
```diff
@@ -0,0 +1,64 @@
+import { mount } from 'enzyme';
```
marbemac: Can we move this test into `src/hooks/__tests__`? Then good to merge!

PR #14, opened 2018-11-13
SL-144/portal

P0lip:
I am not quite sure if it makes sense to include a story, as in fact, it doesn't have any fancy styling.
Moreover, it gets rendered in a different place. I added a story just for documentation purposes.

I decided to write some basic test cases. I tweaked our Jest configuration a bit.


src/Portal.tsx
```diff
@@ -0,0 +1,34 @@
...
+}
+
+export class Portal extends PureComponent<IPortalProps> {
+  private readonly el!: HTMLDivElement;
+  private readonly root = document.body;
+
+  constructor(props: IPortalProps) {
+    super(props);
+
+    this.el = document.createElement('div');
```
marbemac: While not a strict requirement right now, I'd generally suggest protecting browser global usage in react components (for SSR). Something as simple as 

```js
if (typeof document === 'undefined') return;
```

Of course that means we should update `private readonly el!: HTMLDivElement;` to `private readonly el?: HTMLDivElement;` and add `if (!this.el) return null;` where appropriate.

src/__tests__/Portal.spec.tsx
```diff
@@ -0,0 +1,95 @@
...
+  });
+
+  it('attaches className when given', () => {
+    const className = 'test-classname';
+    shallow(<Portal className={className}>test</Portal>);
+
+    expect(setClassNameSpy).toHaveBeenCalledWith(className);
+  });
+
+  describe('lives in non-browser environment', () => {
```
P0lip: I am not sure if we want to have such cases tested, as TS has us covered to some extent.
If so, it might be a good idea to split such tests, so that we don't have to manually mock/unmock the environment.
@marbemac @casserni WDYT guys?

PR #16, opened 2018-11-14
Sl 640/stateful-components

casserni:
General idea here is to allow components to be either controlled or uncontrolled, setting them up this way allows for more flexibility later down the line and also allows us to control the inputs/outputs of native onchange, etc.. functions


marbemac:
Looks good to me, but suggest @P0lip reviews and approves.

Side note, I want to use functional components + hooks rather than class components and lifecycle methods so badly 🤓. Requires react 16.7, which studio is currently using - thoughts?

src/Textarea.tsx
```diff
@@ -7,12 +8,42 @@ export interface ITextareaProps extends ITextProps {
...
+
+  public render() {
+    const { autosize, minRows, maxRows, ...rest } = this.props;
+    const value = (this.props.hasOwnProperty('value') ? this.props.value : this.state.value) || '';
+
+    if (autosize) {
+      return <AutosizeTextarea {...rest} minRows={minRows} maxRows={maxRows} value={value} onChange={this.onChange} />;
+    }
+
+    return React.createElement('textarea', { ...rest, value, onChange: this.onChange });
```
P0lip: I'd stick to JSX here :)

```js
return (
  <textarea
    {...rest}
    value={value}
    onChange={this.onChange}
  />
);
```

stories/Textarea.tsx
```diff
@@ -49,9 +55,24 @@ export const textareaAutosizeKnobs = (tabName = 'Textarea'): any => {
...
+  .add('uncontrolled', () => <Textarea {...textareaKnobs()} {...textKnobs()} {...boxKnobs()} />)
+  .addDecorator(StateDecorator(store))
   .add('with defaults', () => (
-    <Textarea {...textareaKnobs()} {...textKnobs()} {...boxKnobs()} defaultValue="Textarea Text" />
+    <Textarea
+      {...textareaKnobs()}
+      {...textKnobs()}
+      {...boxKnobs()}
+      value={store.get('value')}
+      onChange={(value: any) => store.set({ value })}
```
P0lip: I know it's a story, but we could change the type here to string since we know it.

src/Textarea.tsx
```diff
@@ -7,12 +8,42 @@ export interface ITextareaProps extends ITextProps {
...
+export class BasicTextArea extends React.Component<ITextareaProps, ITextareaState> {
+  constructor(props: ITextareaProps) {
+    super(props);
+    this.state = { value: this.props.value };
+  }
+
+  private onChange = (event: React.ChangeEvent<HTMLTextAreaElement>) => {
+    this.setState({ value: event.target.value });
+
+    if (this.props.onChange) {
```
P0lip: Consider using a default prop here.

src/Textarea.tsx
casserni: Updated!

src/Textarea.tsx
marbemac: Might mean defaultProps (like default noop onChange?). I think either way is good, will leave it to ya'll.

src/Textarea.tsx
```diff
@@ -7,12 +9,32 @@ export interface ITextareaProps extends ITextProps {
...
 }
 
-export const Textarea = styled<ITextareaProps>(Text as any).attrs({
-  as: ({ autosize }: ITextareaProps) => (autosize ? AutosizeTextarea : 'textarea'),
-})(
+export const BasicTextArea = (props: ITextareaProps) => {
+  const { autosize, minRows, maxRows, onChange = noop, ...rest } = props;
+
+  const [value, setValue] = React.useState(props.value || '');
+  const internalValue = props.hasOwnProperty('value') ? props.value : value;
```
marbemac: Instead of managing this extra value, might be able to leverage `useEffect` in a clever way, as below. Note, note 100% on the re-render logic, would need to test it out.

```tsx
export const BasicTextArea = (props: ITextareaProps) => {
  const { autosize, minRows, maxRows, onChange = noop, ...rest } = props;

  const [value, setValue] = React.useState(props.value || '');
  
  React.useEffect(() => {
    if (value !== props.value) setValue(props.value);
  }, [props.value])

  const handleChange = (event: React.ChangeEvent<HTMLTextAreaElement>) => {
    setValue(event.target.value);
    onChange(event.target.value);
  };

  if (autosize) {
    return (
      <AutosizeTextarea {...rest} minRows={minRows} maxRows={maxRows} value={value} onChange={handleChange} />
    );
  }

  return <textarea {...rest} value={value} onChange={handleChange} />;
};
```

If that works, then we can create a sweet little re-usable controllable value hook, as below.

```tsx
export const useControllableValue = (externalValue) => {
  const [value, setValue] = React.useState(externalValue || '');
  
  React.useEffect(() => {
    if (value !== externalValue) setValue(externalValue);
  }, [externalValue])

  return [value, setValue];
}

export const BasicTextArea = (props: ITextareaProps) => {
  const { autosize, minRows, maxRows, onChange = noop, ...rest } = props;

  const [value, setValue] = useControllableValue(props.value);
  
  const handleChange = (event: React.ChangeEvent<HTMLTextAreaElement>) => {
    setValue(event.target.value);
    onChange(event.target.value);
  };

  if (autosize) {
    return (
      <AutosizeTextarea {...rest} minRows={minRows} maxRows={maxRows} value={value} onChange={handleChange} />
    );
  }

  return <textarea {...rest} value={value} onChange={handleChange} />;
};
```

src/Textarea.tsx
P0lip: Yeah, I mean defaultProps, such as noop onChange, but I am fine with both.

src/Textarea.tsx
P0lip: @marbemac @casserni 
extracted that controllableValue here https://github.com/stoplightio/ui-kit/pull/15/commits/b8a7da108ba429138eba78658baa7e778677f6aa
Working flawlessly.

casserni: @P0lip Solid! Once Code-editor is merged in I will go back and update these components! Merging for now

PR #18, opened 2018-11-16
SL-652/menu component

P0lip:


src/ContextMenu.tsx
```diff
@@ -11,6 +11,10 @@ import { ITextProps, Text } from './Text';
 import { styled } from './utils';
 
 // TODO: expose SubMenu component
+// TODO: allow custom renderMenu
+// TODO: allow custom renderMenuItem?
+// TODO: add icon support to menu items
+// TODO: Context Menu should probably leverage/use MENU
```
casserni: Thought here is that contextMenu should use menu, we want them to look the same/have similar functionality except one opens on right click and the other shows on click/hover

src/Menu.tsx
```diff
@@ -0,0 +1,82 @@
...
+// TODO allow custom renderMenuItem
+// TODO allow custom renderTrigger to make this a popup type component
+
+/**
+ * MENU
+ */
+export interface IMenuProps {
+  menuItems: IMenuItemProps[];
+
+  attributes?: IBoxProps;
```
casserni: Try to avoid used the styled util when possible. I tried something new with menu which I think could be a good convention. More "complex" components that use box as a base should take an attributes prop which can be used to override the default styling applied directly to box. This makes it a little cleaner by having us define the style directly in the component, and also separates the overridable style props from the important component ones like menuItems

src/Menu.tsx
```diff
@@ -0,0 +1,82 @@
+import * as React from 'react';
+
+import { Box, IBoxProps } from './Box';
+import { Flex } from './Flex';
+import { Icon, IIcon } from './Icon';
+
+// TODO allow dividers in the menu
+// TODO allow custom renderMenu
+// TODO allow custom renderMenuItem
+// TODO allow custom renderTrigger to make this a popup type component
```
casserni: Good start! Updated Menu to take an items props rather than freeform children so that we can enforce consistency. Eventually we want to update to support custom render functions, I added some todo comments as reminders. Another thing we probably want to support is a renderTrigger so that menus could work as popups/dropdowns, but this should be in another story!

PR #21, opened 2018-11-16
WIP Architecture Changes

casserni:
- [x] update rules and components so that `@...` points to theme values


src/__stories__/_utils.ts
```diff
@@ -3,44 +3,44 @@
...
 export const TextAlign = ['', 'left', 'right', 'center', 'justify', 'initial', 'inherit'];
 export const OverFlow = ['', 'visible', 'hidden', 'scroll', 'auto', 'initial', 'inherit'];
 export const PositionOpts = ['', 'static', 'relative', 'fixed', 'absolute', 'sticky'];
 
-export const FontSize = ['', 'xs', 'sm', 'md', 'lg', 'xl', '2xl', '3xl', '4xl', '5xl', '6xl'];
-export const FontWeight = ['', 'thin', 'light', 'normal', 'medium', 'semibold', 'bold', 'extrabold'];
+export const FontSize = ['', '@xs', '@sm', '@md', '@lg', '@xl', '@2xl', '@3xl', '@4xl', '@5xl', '@6xl'];
+export const FontWeight = ['', '@thin', '@light', '@normal', '@medium', '@semibold', '@bold', '@extrabold'];
 
 export const Size: SizeProp[] = ['xs', 'lg', 'sm', '1x', '2x', '3x', '4x', '5x', '6x', '7x', '8x', '9x', '10x'];
```
P0lip: Aren't we missing `@` here?

src/utils/rules.ts
```diff
@@ -66,6 +71,42 @@ export const styleColor = ({ prop, cssProperty }: { prop: string; cssProperty?:
...
+  prop: string;
+  cssProperty?: string;
+  key?: string;
+  transformValue?: Function;
+  // TODO FIXME
+}): TemplateStringsArray => {
+  const cssName = cssProperty || prop;
+  const fn = (props: any) => {
+    let val = props[prop];
+    if (!is(val)) return null;
```
P0lip: We have a similar functionality in `styleColor`. Could we reuse it or extract to some shared util function?

src/__stories__/_utils.ts
casserni: Size here refers to the FontAwesome sizes, which are not defined in the theme so don't need the `@`

src/utils/rules.ts
casserni: Similar but different, in `styleColor` there is some recursive action to resolve json refs in the theme so that colors can reference eachother. Although that functionality might make sense to have always.  Will see if there is a smart way to handle!

marbemac:
Quick note that I'll have time to review this manana!

src/utils/rules.ts
```diff
@@ -38,21 +39,25 @@ export const styleColor = ({ prop, cssProperty }: { prop: string; cssProperty?:
       val = get(props.theme, val);
     } else {
```
P0lip: I think we could merge that else.

`else if (val.startsWith('@')) {`

src/utils/rules.ts
```diff
@@ -38,21 +39,25 @@ export const styleColor = ({ prop, cssProperty }: { prop: string; cssProperty?:
...
       // is taken from a prop
-      if (typeof val === 'string') {
-        val = val.split('.');
-      }
+      if (val.startsWith('@')) {
+        if (typeof val === 'string') {
```
P0lip: This condition here doesn't make much sense to me, since at line 42 `val.startsWith('@')` we already assume that val is indeed a string. If we are not certain whether val is a string, let's move that condition above to avoid TypeErrors.

src/utils/rules.ts
```diff
@@ -250,12 +296,18 @@ const getSpaceValue = scale => propVal => {
 // @ts-ignore FIXME
 export const space = props => {
   // test for spacing props, so m* and p*
-  const keys = Object.keys(props).filter(key => /^[mp][trblxy]?$/.test(key));
   // get space configuration from theme
   const scale = get(props.theme, 'base.space');
   // function to convert propVal -> cssVal
   const getStyle = getSpaceValue(scale);
 
+  // to ensure proper order
```
P0lip: Would be nice to extract the keys and sort them somehow, but looking at the code we may need to supply the order list manually.

```js
const order = ['m', 'p', 'x', 'y', 't', 'r', 'b', 'l'];
const keys = Object.keys(props)
  .filter(key => /^[mp][trblxy]?$/.test(key))
  .sort((keyA, keyB) => order.indexOf(keyA[keyA.length - 1]) - order.indexOf(keyB[keyB.length - 1]));
```
It doesn't look super clean, though, so I leave it to you.

src/types.ts
```diff
@@ -34,26 +114,27 @@ export interface ISectionTheme<TComponents extends Components = Components> {
 }
 
 export interface ILayout {
-  textSize: { [key in FontSize]?: number | string }; // px
+  textSize: { [key in keyof IFontSize]?: number | string }; // px
```
rainum: 👍 for the `keyof` keyword

src/utils/rules.ts
```diff
@@ -13,13 +13,14 @@ const propThemeMap = {
...
 
+// TODO FIX
 export const styleColor = ({ prop, cssProperty }: { prop: string; cssProperty?: string }) => {
   const resolvedProp = cssProperty || prop;
 
   // create a cache to check and break circular references
   // @ts-ignore FIXME
   const fn = (props: any, cache: any[] = []) => {
-    let val = props[prop];
+    let val = props[prop] || '';
```
rainum: There is no need in a `''`, since if `val` is falsy function will always return a `null`.

```suggestion
    let val = props[prop];
```

src/utils/rules.ts
rainum: yeah they're too different to combine them

src/utils/rules.ts
```diff
@@ -240,7 +286,7 @@ const getSpaceValue = scale => propVal => {
     }
 
     // check the theme config for a value, or just use the prop
-    val = scale[val] || val;
+    val = val.startsWith('@') ? scale[val.slice(1)] : val || val;
```
rainum: That looks confusing 😮 Better to add some brackets here

```suggestion
    val = (val.startsWith('@') ? scale[val.slice(1)] : val) || val;
```

marbemac:
@casserni can you review the outstanding comments and either reply and/or resolve? Will help to move it along!

Also, should this or the types PR be reviewed/merged first?

src/utils/rules.ts
casserni: Updated!

src/utils/rules.ts
casserni: Updated!

src/utils/rules.ts
casserni: Leaving as is for readability can go back later to restructure

src/utils/rules.ts
casserni: Found a solution to combine!

casserni:
@marbemac The PR should be merged before typings PR, this one has been updated some and can use another look

src/Box.tsx
```diff
@@ -30,50 +29,59 @@ import {
...
-  minWidth?: BoxDimension | string | number;
-
-  border?: BorderWidth;
-  borderTop?: BorderWidth;
-  borderLeft?: BorderWidth;
-  borderRight?: BorderWidth;
-  borderBottom?: BorderWidth;
+  weight?: ValueOf<IFontWeight>;
+
+  m?: ValueOf<IFullSpace>;
```
marbemac: Can this be consolidated to something like `type SpacingVal = ValueOf<IFullSpace>` - and then we use SpacingVal everywhere needed?

src/Box.tsx
```diff
@@ -30,50 +29,59 @@ import {
...
+  pb?: ValueOf<IFullSpace>;
+  pl?: ValueOf<IFullSpace>;
+  px?: ValueOf<IFullSpace>;
+  py?: ValueOf<IFullSpace>;
+
+  height?: ValueOf<IBoxDimension> | string | number;
+  maxHeight?: ValueOf<IBoxDimension> | string | number;
+  minHeight?: ValueOf<IBoxDimension> | string | number;
+  width?: ValueOf<IBoxDimension> | string | number;
+  maxWidth?: ValueOf<IBoxDimension> | string | number;
```
marbemac: Same as above SpacingVal comment but for this situation. Same anywhere else that we repeat the same typing more than like 5 times or something.

src/Break.tsx
```diff
@@ -17,6 +17,6 @@ export const Break = styled<IBreakProps, 'hr'>(Box as any)(thickness);
 Break.defaultProps = {
   as: 'hr',
   thickness: 1,
-  m: 'none',
-  borderColor: 'break.border',
+  m: '@none',
```
marbemac: Don't like this. `@none` should be `0` in the case of margin, but `none` in the case of border or shadow etc. Thus, this is ambiguous. Perhaps just `0`? Do we need theme variables representing none at all?

src/Checkbox.tsx
```diff
@@ -39,12 +39,12 @@ export const BasicCheckbox = (props: ICheckboxProps) => {
...
+        m="@none"
+        p="@none"
+        radius="@md"
         items="center"
         justify="center"
-        bg={isChecked ? 'toggle.checked.bg' : 'toggle.bg'}
+        bg={isChecked ? '@toggle.checked.bg' : '@toggle.bg'}
         cursor={disabled ? 'not-allowed' : 'pointer'}
         width={width || '20px'}
         height={height || '20px'}
```
marbemac: I spy hardcoded literal values.

src/Table.tsx
```diff
@@ -43,27 +43,27 @@ export const Table = styled<ITableProps, 'table'>(Box as any).attrs({
...
 );
 
 export const TableCell = styled<ITableCellProps, 'td'>(Box as any).attrs({
   as: 'td',
-  border: 'none',
-  borderLeft: 'sm',
-  borderColor: 'colors.border',
+  border: '@none',
+  borderLeft: '@sm',
+  borderColor: '@border',
```
marbemac: Should these be `@table.border` to be consistent with all the other components?

src/ThemeProvider.tsx
```diff
@@ -0,0 +1,33 @@
...
+    if (typeof val === 'string' && val.startsWith('#/')) return val;
+    return;
+  },
+});
+
+export interface IThemeProviderProps {
+  theme?: IThemeInterface;
+  children?: any;
+}
+export const ThemeProvider = (props: IThemeProviderProps) => {
```
marbemac: Suggest using state + only run resolver on mount/dismount for now. COULD pass in theme object to run resolver when it changes, not sure if needed though. 

Note, useEffect cannot be an async function, so I define the function outside of the effect (which is harmless).

Example:

```tsx
export const ThemeProvider = (props: IThemeProviderProps) => {
  const [theme, setTheme] = useState(Object.assign({}, baseTheme, props.theme));

  const resolveTheme = async () => {
    const resolved = await resolver.resolve(theme);
    
    if (resolved.errors && resolved.errors.length > 0) {
      console.error('Error resolving theme object.', resolved.errors)
      return;
    }

    setTheme(resolved.result);
  }

  React.useEffect(() => {
    resolveTheme();
  }, []);

  return <StyledProvider theme={theme}>{props.children}</StyledProvider>;
};
```

src/ThemeSection.tsx
```diff
@@ -1,24 +1,24 @@
...
 interface IThemeSectionProps {
-  theme?: ISectionTheme;
+  theme?: IThemeInterface;
+  sectionTheme?: ISectionTheme;
   section: string;
   children?: any;
 }
 
-export class ThemeSection extends React.Component<IThemeSectionProps> {
+export class ThemeSectionBase extends React.Component<IThemeSectionProps> {
```
marbemac: Feels like a good hooks candidate - I'll update push something :).

marbemac:
Looks promising overall - added some comments.

Still not completely happy with how themes are defined, and the mixing of traditional themes (colors) and structure (margin values, etc). However, that's out of scope of this PR - will think on it.

src/ThemeProvider.tsx
casserni: If we only run on mount/unmount the ui won't get appropriately updated with the theme object changes

P0lip: A bit out of context, but that callback could be omitted I suppose, so ultimately the effect hook would be invoked in the following way:
```js
React.useEffect(resolveTheme, [props.theme]);
```

marbemac: I think that `useEffect` only supports a function that returns `void | () => void` (disposer). I don't think it supports returning a promise (which is what resolveTheme does).

casserni:
On hold while we try out a new styling architecture

PR #23, opened 2018-11-19
Sl 678/typings

casserni:
needs https://github.com/stoplightio/ui-kit/pull/21


casserni:
On hold while we try out a new styling architecture

PR #28, opened 2018-11-23
SL-701: build(package): limit the scope of files in published package

chris-miaskowski:
we included all files in the folder, this could potentially cause leaking passwords or other sensitive data

SL-632

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
   },
   "license": "Apache-2.0",
   "files": [
-    "**/*"
+    "dist/**/*"
   ],
   "engines": {
     "node": ">=8.3.0"


P0lip:
I believe sl-scripts' release command takes care of that.
@marbemac could you confirm?

I cannot seem to find any irrelevant files (such as CircleCI configs) in the final bundle that is uploaded to npm. At the moment, we seem to pack all files under `src` directory, LICENSE, README.md and obviously package.json.

marbemac:
Yeah this should be taken care of already.

What happens is the build process compiles all the `/src` files into the `/dist` folder. Then a few other important files are copied over to dist (`package.json`, `license`, `readme.md`). Finally, we `cd` into `dist` and publish from that directory. Thus, only built files + the three extra files described above are ever published.

chris-miaskowski:
Closing as irrelevant, thanks for comments everyone!

PR #30, opened 2018-11-27
fix(code-editor): fail gracefully when cannot highlight code SL-715

chris-miaskowski:
[SL-715](https://stoplightio.atlassian.net/browse/SL-715)

**Notes for reviewer**

- `prismjs` loads only a handful of syntax highlighters by default (marked as `defaultLanguages` in the code)
- slate-editor claims to support all other languages but since prismjs does not load them be default, switching to one of them fails with an exception:

selection_022

- my code lists and groups all supported languages but comments them out because they won't load out of the box. we will have to figure out a proper babel configuration to load them (https://github.com/mAAdhaTTah/babel-plugin-prismjs).

### DIFF
--- a/src/CodeEditor.tsx
+++ b/src/CodeEditor.tsx
@@ -147,3 +147,7 @@ export const CodeEditor = styled<ICodeEditorProps>(CodeEditorView as any)`
     }
   }
 `;
+
+CodeEditor.defaultProps = {
+  language: defaultLanguage,
+};
--- a/src/CodeEditor.tsx
+++ b/src/CodeEditor.tsx
@@ -13,6 +13,44 @@ export interface ICodeEditorProps {
   onChange?: (code: string) => any;
 }
 
+const defaultSupport = {
+  css: 'CSS',
+  javascript: 'JavaScript',
+  markup: 'Markup',
+  clike: 'C-like',
+};
+
+const optionalSupport = {
+  // cpp: 'C++',
+  // csharp: 'C#',
+  // go: 'Go',
+  // html: 'HTML',
+  // http: 'HTTP',
+  // java: 'Java',
+  // json: 'JSON',
+  // jsx: 'JSX',
+  // markdown: 'Markdown',
+  // objectivec: 'Objective-C',
+  // perl: 'Perl',
+  // php: 'PHP',
+  // python: 'Python',
+  // ruby: 'Ruby',
+  // scala: 'Scala',
+  // bash: 'Bash',
+  // swift: 'Swift',
+  // yaml: 'YAML',
+};
+
+/**
+ * List of supported languages: https://prismjs.com/#languages-list
+ */
+export const supportedLanguages = {
+  ...defaultSupport,
+  ...optionalSupport,
+};
+
+export const defaultLanguage = 'javascript';
+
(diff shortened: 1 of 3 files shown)


src/CodeEditor.tsx
```diff
@@ -13,6 +13,44 @@ export interface ICodeEditorProps {
...
+  // scala: 'Scala',
+  // bash: 'Bash',
+  // swift: 'Swift',
+  // yaml: 'YAML',
+};
+
+/**
+ * List of supported languages: https://prismjs.com/#languages-list
+ */
+export const supportedLanguages = {
```
P0lip: Any point of exporting this info? I mean, we import all available languages at the moment, therefore this object contains incorrect information.

src/CodeEditor.tsx
```diff
@@ -13,6 +13,44 @@ export interface ICodeEditorProps {
...
+  ...defaultSupport,
+  ...optionalSupport,
+};
+
+export const defaultLanguage = 'javascript';
```
P0lip: Could be a default prop.

src/CodeEditor.tsx
chris-miaskowski: I will use that list in `slate-editor` to build a dropdown (no point duplicating this in slate-editor repo)

src/CodeEditor.tsx
chris-miaskowski: Not sure I get this. It's not a default export :) it's a hint for users of CodeEditor component which language to fallback by default

P0lip: I mean it could be default prop for the CodeEditor component, so that when you don't pass any language, then the default one is used.

src/CodeEditor.tsx
P0lip: What I mean is that the list of supported languages is incorrect. We should export everything that's available. If you need that only for Slate, define the list in Slate.

src/CodeEditor.tsx
chris-miaskowski: ooh :) Ok good point, but the exported value can be useful too (for selecting a default value in a dropdown for instance)

src/CodeEditor.tsx
chris-miaskowski: What do you mean that it's incorrect? It is exactly what the CodeEditor will support right now. It fails on any language than 
```
  css: 'CSS',
  javascript: 'JavaScript',
  markup: 'Markup',
  clike: 'C-like',
```

src/CodeEditor.tsx
P0lip: Yeah, I am fine with exporting it :) Just wanted to have a default one too.
Make sure to adjust ICodeEditorProps to reflect the default prop.

src/CodeEditor.tsx
chris-miaskowski: I don't see why Slate would define what is supported. It's not Slate that defines that but the CodeEditor. What if Slate 'tells' that "hungarian" is a supported language? But CodeEditor clearly doesn't support that :)

P0lip: Ah yeah, thought we imported everything. It's ok then, let's leave it as is.

PR #31, opened 2018-11-27
SL-742 use single Lodash imports

P0lip:


wmhilton:
Oooh, does `import get from 'lodash/get'` not work? We have to use require?

marbemac:
@wmhilton yeahh, lodash typings are messed up.

P0lip:
Yep, sa Marc said lodash typings are messed up.

wmhilton:
Big win!
image

PR #34, opened 2018-11-28
SL-639: forward ref to textarea

wmhilton:
To do some of the fancy keyboard navigation needed for the Slate editor, we need access to the underlying textarea component, so I am forwarding a ref.


src/CodeEditor.tsx
```diff
@@ -49,17 +49,17 @@ export const supportedLanguages = {
   ...optionalSupport,
 };
 
-const CodeEditorView = (props: ICodeEditorProps & { className: string }) => {
+const CodeEditorView = React.forwardRef((props: ICodeEditorProps & { className: string }, ref: any) => {
```
P0lip: Let's add some typing here.
```ts
React.forwardRef<ReactSimpleCodeEditor, ICodeEditorProps & { className: string }>((props, ref) => {
```

wmhilton: Why? TS seems to do a good job infering it automatically.
image

Edit: Ooooh I see, it _doesnt_ do a good job inferring it. `ForwardRefExoticComponent` is not as specific as `ReactSimpleCodeEditor`. OK, I got this!

wmhilton: Weirdly, it seems like TS still sees the same type:
image

P0lip: It does not seem to be exactly the same one. See the lack of ReactSimpleCodeEditor in RefAttributes in the first screenshot.

wmhilton: I'm nearsighted... I only see the first 3 lines of any TypeScript type. :)

PR #36, opened 2018-11-30
refactor: emotion + new themeing approach + new typings approach

marbemac:
Very much a work in progress! Opening PR for early feedback.

Goal with this entire thing is to make the code simpler, typings better, and stronger more clear theming approach.

- Separate structure (spacing, flex, fonts, etc) from theme (colors).
- Separate "zones" (sections) from "theme". Note that both can still be customized by the library user.
- "baseTheme" concept (dark or light). The library user's custom theme simply extends one of those base themes (similar to vscode's theming approach).
- One story that shows some random usage combinations (the first one, Layout).
- Box, Flex, and Button migrated (components in the `emotion` folder).

Note, have not integrated the storybook theme buttons yet.


src/emotion/Button.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { jsx } from '@emotion/core';
+import { HTMLAttributes, SFC } from 'react';
+
+import { Box, IBox } from './Box';
+import { useTheme } from './theme';
+
+export const Button: SFC<IButton> = props => {
+  const { as = Box, disabled, ...rest } = props;
+
+  const css = [...buttonStyles({ disabled })];
```
P0lip: Custom styles are ignored here I guess :)

marbemac: Nope! Because it just gets passed through to our foundational component (Box).

casserni: @marbemac assuming that the user doesn't change the `as` prop and do something like
`<Button as="div" />`

src/emotion/Button.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  return jsx(as, {
+    ...rest,
+    as: 'button',
+    css,
+  });
+};
```
casserni: related to above I think it makes more sense to do 
  
```
return jsx(Box, {
    ...rest,
    as: props.as || 'button',
    css,
  });
```

src/emotion/Flex.tsx
```diff
@@ -0,0 +1,41 @@
...
+    css,
+  });
+};
+
+export interface IFlex
+  extends IBox,
+    ss.FlexBasisProps,
+    ss.FlexDirectionProps,
+    ss.FlexWrapProps,
+    ss.AlignItemsProps,
```
casserni: Although the interfaces look clean `AlignItemsProps` is an example of where autocomplete wont work since the value is like below

```
export type AlignItemsProperty = 
   | Globals 
   | SelfPosition 
   | "baseline" 
   | "normal" 
   | "stretch" 
   | string;
```

src/emotion/theme.tsx
```diff
@@ -0,0 +1,135 @@
...
+/**
+ * Theme
+ */
+
+interface ITheme {
+  base: BaseTheme;
+
+  canvas: {
+    fg: string;
+    bg: string;
```
casserni: How to we feel about being more explicit here? So instead of `bg` and `fg` we change to `background` and `foreground`, makes it more clear as an outside user looking to what they mean

src/emotion/Flex.tsx
marbemac: Good point, will fix - I have an idea.

src/emotion/theme.tsx
marbemac: Medium - what do others think?

If we're not nesting, then things like `selectedItemBg` would become pretty long: `selectedItemBackground`.

lottamus: I'm good either way as long as it auto-completes. In studio we should have a visual layer on top like with linting.

lottamus: I'm good either way as long as it auto-completes. In studio we should have a visual layer on top like with linting.

marbemac: Toss up to me - I'm going to leave as is, think it's fine.

marbemac:
@P0lip @casserni @lottamus ok so in conclusion, what do you guys think? To me, this approach feels pretty solid.

If we're all into it and @P0lip you find yourself with some free time, would you be interested in taking a stab at converting some of the other components?

src/emotion/Button.tsx
```diff
@@ -0,0 +1,62 @@
...
+  const css = [...buttonStyles({ disabled })];
+
+  return jsx(Box, {
+    ...rest,
+    as,
+    css,
+  });
+};
+
+export interface IButton extends IButtonProps, IBox, HTMLAttributes<'button'> {}
```
P0lip: Let's have a generics here - 'button' should be a default, but not hardcoded.
Why? `As` is not enforced and can be changed.

P0lip: Looking at signatures, I think we have to use HTMLButtonElement here. It's primarily used in SyntheticEvent, so I think we can live with button here.

marbemac: Agreed on both! And as a general pattern in all the components (Box, Flex, etc).

marbemac:
> The second issue I faced is that at the very moment Emotion-based components don't resolve variables from the theme.

@P0lip do you mean from the Storybook addon theme selector? If so, then yeah I have not hooked up the storybook theme addon we made to the new theme approach - it is still built on / connected to the styled components theming.

> The current theme structure doesn't really allow us to exchange some non-color values

Yup, we should expose a way to do this, but I think it should be separate from the theming. Will tackle this tonight.

marbemac:
Ah, I intentionally ommitted support for that pattern. It was introducing all sorts of complexity into ui-kit and usage patterns (adding the @ symbol to distinguish "red" color from "@red" variable, etc). The new pattern is to just access the well typed theme :).

Instead of this:

```jsx
export const MyComponent = () => {
  return (
    <Box bg="@red" />
  );
}
```

You'd write:

```jsx
export const MyComponent = () => {
  const theme = useTheme();
  return (
    <Box bg={theme.red} />
  );
}
```

Much more explicit, we don't have to create the whole @sigil string lookup / merging crazy system, and it fully supports typings.

P0lip:
Gotcha! Makes sense to me, as passing a raw value is indeed a cleaner and preferred way.
I will adjust the rest of the components then, we are very likely to proceed with this new approach I guess.
I'll spin off a second PR for that just not to mess up the current one unless you are fine with me working on the same branch :D

marbemac:
Great, let's do it! 🎉 

Feel free to continue on this PR, np.

marbemac:
Ok I pushed a few changes. 

- Moved all styled-components stuff to `__old__` folder. We will delete this before merging PR.
- Moved all emotion components root src.
- Hooked up our custom storybook theme addon to the new theme system - the light/dark toggle works now.

Regarding customizing layout properties (spacing, etc), I suggest we just render the `ThemeProvider` from `styled-system` (which we are using under the hood) if we need to customize the spacing: https://github.com/jxnblk/styled-system/blob/master/docs/getting-started.md#space-theming - simple :).

Thought re Storybook usability below:

---

@P0lip I think we can also take this opportunity to improve the Storybook. Right now, we have a single story per component, which I think is not the correct approach. I'd encourage us to build the stories around common use cases and "questions" end developers might have. What would YOU want to see in the docs? Some examples of how others organize their docs:

- https://evergreen.segment.com/components
- https://blueprintjs.com/docs/#core/components/file-input
- https://react.semantic-ui.com/

Our "groups" might look something like the below, which use case stories in each group.

- Layout
- Buttons
- Forms
- Toggles
- Menus
- Feedback
- Lists & Tables
- Overlays
- Utilities & Helpers

In conclusion, as we move stories over I'd encourage us to improve the usefulness of the Storybook.

P0lip:
@marbemac converted, polished and cleaned up a few stories to match the new approach.
I kinda like it! I suppose we may want to add some more examples to showcase all capabilities rather than just relying on knobs.

Regarding spacing - it should be sufficient, but how should we deal with other non-spacing properties such as borders? They are unlikely to be left unresolved. We can include them in the theme, though.

P0lip:
@marbemac 
I finished the migration (all components are moved excluding Theme*).
Some of them may need slight adjustments, as paddings/margins have been altered.

There are a couple of questions left to be answered:

1. createElement vs JSX? Right now we have mixed usage.

2. Dealing with refs.
Function components cannot be given refs. I had some issues with that I needed to wrap some components with plain nodes.

3. Nested components styling, i.e. ContextMenu/Menu.
It would be nice to have some consistent way of styling for more complex components.
Also, we don't have a way to pass attributes to such components. 
I'd stick to render props, but made them more granular. 
A top-level component would receive all properties.

marbemac:
> They are unlikely to be left unresolved. We can include them in the theme, though.

I think we were over thinking things before - in the new world its just back to basics, which is much simpler to understand and support in ui-kit:

```
<Box border={`3px solid ${theme.border}`} />
```

> createElement vs JSX? Right now we have mixed usage.

Should use the pragma jsx from emotion everywhere we can.

> Function components cannot be given refs.

Are you sure? Could you link to an example in the branch that shows the issue? See something like https://github.com/stoplightio/monaco/blob/master/src/CodeEditor.tsx#L34

> Nested components styling

Does this feature create any ideas, would it help? If so, we could come up with a standard pattern so that component styles could target other components or something. https://emotion.sh/docs/styled#targeting-another-emotion-component

> tests

:) I'm open to suggestions - what do you think would be best?

P0lip:
```
I think we were over thinking things before - in the new world its just back to basics, which is much simpler to understand and support in ui-kit. Border width is much less important / variable than something like spacing, so inlining it is just fine:
``` 
Sounds fine to me. Just wanted to make sure we are on the same page since plenty of components used some shared values. I've already changed that for almost all new components.

```Should use the pragma jsx from emotion everywhere we can.```
Alright, will change in a few places.

Regarding refs issue

`Function components cannot be given refs.` 
Wasn't clear enough with that one. You are right function components can be given refs. We even do that in case of Code Editor. What I meant is that our 'base' function components such as Box don't seem to propagate the ref properly. I will fix that after daily :)

Here is such a place https://github.com/stoplightio/ui-kit/blob/1b47feebb5edabfe9168e1f7ec0eeb7419bacd39/src/ScrollBox.tsx#L102L105


`:) I'm open to suggestions - what do you think would be best?`
In case of most components - I would just test conditional styles + see if they propagate props to Box/Flex/Text. If they have some logic, such as controlled mode or something similar, we should test it either.

**Comment by marbemac** [comment]
That all sounds good! Simply remove WIP label when you think it's ready, and add nick and I to review.

For the ref thing, https://reactjs.org/docs/forwarding-refs.html could help?

**Comment by P0lip** [comment]
@marbemac @casserni @lottamus @rainum 
Let's go ahead and review the PR. 
In the meantime, I will write a few more tests to match previous coverage level, but other than that we should be all set.

File: src/Table.tsx [hunk 6]
```diff
@@ -1,70 +1,104 @@
...
 
-interface ITableViewProps {
-  className: string;
-  children: any;
+export const tableStyles = ({ isSelected }: ITableProps) => {
+  const theme = useTheme();
+
+  return [
+    {
+      border: '0',
```
rainum: Better is to use:

```suggestion
      border: '0 none',
```

src/Table.tsx
```diff
@@ -1,70 +1,104 @@
...
+};
+
+export interface ITableRow extends IBox<HTMLTableRowElement> {}
+
+export const tableRowStyles = () => {
+  const theme = useTheme();
+
+  return [
+    {
+      border: '0',
```
rainum: Better is to use:

```suggestion
      border: '0 none',
```

src/Table.tsx
```diff
@@ -1,70 +1,104 @@
...
-})(
-  // @ts-ignore
-  boxShadow
-);
+export const tableCellStyles = ({ isSelected }: ITableCellProps) => {
+  const theme = useTheme();
+
+  return [
+    {
+      border: '0',
```
rainum: Better is to use:

```suggestion
      border: '0 none',
```

marbemac:
Looks great! A few small things, then I think we're good to merge.

- [ ] Popup component needs to use new theme system instead of <ThemeConsumer> from styled-components (which I just removed in a recent commit).
- [ ] Storybook having issues?
- [ ] in theme files: `borderColor` -> `border`
- [ ] in theme files: `checkedBorderColor` -> `checkedBorder`
- [ ] in theme files: remove `canvas.link` - what's this for? smells
- [ ] tests seem to have a few "Each child in an array or iterator should have a unique "key" prop" errors

marbemac:
Ok one more after chatting:

- [ ] remove the canvas property from the base theme files, and update any relevant storybooks to just have a `canvas` ThemeZone if needed

marbemac:
@P0lip besides that last remove canvas property bullet in the prior comment, I think this entire PR looks good!

I'm signing off for the night - feel free to approve and merge once the canvas thing is done. Will make it easier for you to work on updating the few repos that depend on the styled-components version of ui-kit to the latest.

P0lip:
@marbemac all issues resolved. Actually, I decided to just rename canvas and totally changed its use.
Apart from that, applied a couple of minor fixes/adjustments to storybook
Merging!

PR #39, opened 2018-12-17
Storybook updates

casserni:
- Removed storybook-addons from the src folder into the .storybook folder
- updated story names so that we have better nesting /ui https://cl.ly/0064a6a2c8f6


.storybook/config.js
```diff
@@ -2,18 +2,24 @@ import { withInfo } from '@storybook/addon-info';
 import { withOptions } from '@storybook/addon-options';
 import { addDecorator, configure } from '@storybook/react';
 
-import * as themes from '../src/storybook-addon/themes';
-import { withThemes } from '../src/storybook-addon/withThemes';
+import * as themes from './custom-addons/withThemes//themes';
```
P0lip: I think we could move this to a separate package since it's used in every single project that requires UI-Kit.
I will do it later.

PR #41, opened 2018-12-19
[SL-156] Storybook documentation

P0lip:
Update the docs to reflect the Emotion change.
Let me know what you think.
I hope I covered everything


casserni:
@P0lip Seeing that Icon commit reminded me, @lottamus brought up a good point that we should avoid using any icons if possible in our base ui-kit components. Since we are leaving it up to the end user to register them, components like checkbox and toggle will not work properly if the icons used don't get registered. Those two components should be easy to rewrite to not include icons, but definitely something to think about as we move forward

wmhilton:
I kinda like that the checkbox component uses a customizable icon!

Maybe you could add a fallback that uses ✅(https://emojipedia.org/white-heavy-check-mark/) if no checkmark icon is registered.

lottamus:
Or we just go ahead and register those font-awesome free icons and add them as dependencies. Then you can override them if you please

wmhilton:
That would work, as long as it's not registering ALL the fontawesome icons (which would add a lot of bundle weight). If it's just checkbox, open folder / closed folder, generic file icon etc that would be awesome to include those by default but have them override-able.

lottamus:
> as long as it's not registering ALL the fontawesome icons

Yeah we would only import and `IconLibrary.add` the ones we need

wmhilton:
Is this PR waiting on anything in particular?

wmhilton:
@lottamus I like that solution, because unlike relying on unicode emojis like ✅ 📁 📂 (which depends on system fonts) it would have a predictable appearance.

P0lip:
@casserni @lottamus @wmhilton 
I second the opinion. I strongly believe we should avoid usage of Icon component in internal UI-Kit components. I don't feel like adding an icon is a good solution for us, as in such scenario we would enforce some icons and that could be confusing.
On the other hand, if we don't do that and someone just registers an icon set, a Toggle (or any other component) component will be messed up, as some of the styles may expect a given shape.
I will just rewrite those components to avoid Icon and pull a plain svg in.

PR #46, opened 2018-12-30
Feature/react window

P0lip:
Good news is Firefox supports [CSS Scrollbars](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Scrollbars) as of 64 version so we should have fairly consistent scrollbar styling for everywhere using plain CSS. The only subtle difference is lack of border-radius.


src/ScrollList.tsx
```diff
@@ -1,102 +1,61 @@
...
   );
 };
+
+export { IFixedSizeList, IVariableSizeList };
+
+export const scrollListStyles = () => {
+  const theme = useTheme();
+
+  return css`
+    &::-webkit-scrollbar {
```
marbemac: How consistent / inconsistent is this with the `ScrollBox` component approach / look and feel?

P0lip: The biggest inconsistency is a lack of autohiding. Scrollbar is always visible unlikes previously where it was visible only during scrolling.
Are we fine with it?

Chromium-based web browsers and Safari should have fairly equal UI. It can be always adjusted since we can treat scrollbar as it was a regular node and most of CSS rules can be applied.

Firefox and others - just limited styling possibilities.

marbemac: Ah interesting. Auto-hide is one of the primary reasons to do custom scrollbars - otherwise we end up with tons of gross scrollbars everywhere. VS code auto-hides - how do they do it if they're doing the css approach?

P0lip: They don't - I was wrong telling you they do, They have plain divs there and just update them as you scroll.
I think we can implement that too even when using react-window, as both `onScroll` and `scrollTo` are exposed.

lottamus:
Should we be memoizing the row before passing it into react-window?

They have an example of doing this in their docs using their `areEqual` function: https://react-window.now.sh/#/examples/list/memoized-list-items

image


Seems to work:

kapture 2019-01-04 at 13 07 28

src/ScrollList.tsx
lottamus: Looks like `react-window` might not support custom scrollbars: https://github.com/bvaughn/react-window/issues/110

I followed up to see if they were able to get it to work or not

P0lip:
Yeah, I use that memoization technique in Tree-List already.

src/ScrollList.tsx
P0lip: Was playing around that and found a solution to achieve autohiding. We can set transparent backgrond-color for scroll. Will finalize this approach.

lottamus:
> Yeah, I use that memoization technique in Tree-List already.

Should we add it to the storybook too?

P0lip:
Sure, will include it.

marbemac:
Removed my review - leaving @lottamus to give final review.

PR #48, opened 2019-01-08
[SL-1150] Dialog component

P0lip:
Decided to start off with a wrapper around Portal

We can extend it later once it turns out all dialogs have some shared layout.

The PR contains a fix for the Button disabled state issue - we didn't pass the prop forward to the component, therefore the button behave as it was enabled (all event handlers were executed)


src/__tests__/Box.spec.tsx
```diff
@@ -13,49 +13,4 @@ describe('Box component', () => {
     expect(ref).toHaveBeenLastCalledWith(expect.any(Object));
     wrapper.unmount();
   });
-
-  it('passes all props but hidden and responsive', () => {
```
P0lip: It must have been erroneously committed one day. Equal tests are present in Image.spec.tsx

PR #50, opened 2019-01-15
feat(contextmenu): SL-1138 adds submenu support to context menu

casserni:
- adds submenu 
- passes trigger ref to render trigger


src/ContextMenu.tsx
```diff
@@ -132,14 +140,46 @@ export const ContextMenuItem: FunctionComponent<IContextMenuItem> = props => {
           data={data}
           preventClose={preventClose}
           disabled={disabled}
-          onClick={onClick}
+          onClick={(e, d, t) => {
```
P0lip: `onClick={onClick}`?

src/ContextMenu.tsx
```diff
@@ -132,14 +140,46 @@ export const ContextMenuItem: FunctionComponent<IContextMenuItem> = props => {
...
       )}
     />
   );
+
+  if (isSubMenu) {
+    return (
+      <Box
+        {...rest}
+        css={menuStyles()}
+        as={({ className }: { className: string }) => {
```
P0lip: I'd move it to a separate component.
Look at Input for a reference.
We could also make use of [Emotion's ClassNames](https://emotion.sh/docs/class-names).

src/ContextMenu.tsx
```diff
@@ -87,6 +87,11 @@ const menuStyles = () => {
       ':focus': {
         outline: '0 none',
       },
+
+      '.react-contextmenu-submenu': {
+        '.react-contextmenu': { opacity: 0, display: 'none' },
```
P0lip: Do we need both `opacity` and `display`?

P0lip: Do we need both `opacity` and `display`?

rainum: Yeah, display is enough in this case.

src/ContextMenu.tsx
```diff
@@ -132,14 +140,46 @@ export const ContextMenuItem: FunctionComponent<IContextMenuItem> = props => {
...
+          onClick={(e, d, t) => {
+            if (onClick) {
+              return onClick(e, d, t);
+            }
+          }}
         >
-          {title && <Text>{title}</Text>}
-          {divider && <Break thickness={1} />}
+          <Flex alignItems="center">
+            {title ? <Text flex={1}>{title}</Text> : null}
```
rainum: Do we really need to use Text here?

```suggestion
            {title ? <Box flex={1}>{title}</Box> : null}
```

src/ContextMenu.tsx
```diff
@@ -132,14 +140,46 @@ export const ContextMenuItem: FunctionComponent<IContextMenuItem> = props => {
...
+            {isSubMenu ? <Text pl="5px">&#9658;</Text> : null}
```
rainum: Same as above

```suggestion
            {isSubMenu ? <Box pl="5px">&#9658;</Box> : null}
```

src/ContextMenu.tsx
casserni: Disagree with moving since it heavily depends on menu item. Cannot currently think of a use case where it doesn't. Also cannot think of a use case where we would need to only import submenu, leaving as it for and can update later if needed

P0lip: Don't remember the case, but I used to have troubles with such an approach.
Basically, in this case, you are kind of creating a new component every single time.
I am fine with leaving it as is given it works as expected.

src/ContextMenu.tsx
```diff
@@ -134,12 +157,40 @@ export const ContextMenuItem: FunctionComponent<IContextMenuItem> = props => {
...
   );
+
+  if (isSubMenu) {
+    return (
+      <Box
+        {...rest}
+        css={menuStyles()}
+        as={({ className }: { className: string }) => {
+          return (
+            // @ts-ignore
```
P0lip: Could we get rid of the comment?

src/__tests__/ContextMenu.spec.tsx
```diff
@@ -39,10 +40,11 @@ describe('ContextMenu component', () => {
     const trigger = 'I am a trigger!';
     const renderTrigger = () => trigger;
 
-    const wrapper = shallow(<ContextMenu renderTrigger={renderTrigger} id="uniq-id" />);
+    const wrapper = mount(<ContextMenu renderTrigger={renderTrigger} id="uniq-id" />);
```
P0lip: Maybe just fake `useRef` implementation? We can get rid of it once shallow rendering is supported/

src/ContextMenu.tsx
```diff
@@ -23,24 +23,38 @@ import { Box, Break, IBox, Text, useTheme } from './';
...
   const { id, renderTrigger, ...rest } = props;
 
+  const contextTriggerRef = React.useRef<any>(null);
+
   return (
     <Fragment>
-      {renderTrigger && <ReactContextMenuTrigger id={id}>{renderTrigger()}</ReactContextMenuTrigger>}
-
+      <ContextMenuTrigger id={id} ref={contextTriggerRef} holdToDisplay={-1}>
+        {renderTrigger(contextTriggerRef)}
```
P0lip: The `contextTriggerRef.current` passed to renderTrigger going to be null initially unless `ContextMenuTrigger` is rendered (which happens after renderTrigger is executed)
Is this intended? I don't think so.

casserni: renderTrigger is now required, `ContextMenuTrigger` should always be rendered

casserni: @P0lip I saw in studio there was an area where we used the trigger and menu separately, but that was done as a workaround for missing functionality (open on clicking the trigger) here. I can't think of a reasonable scenario where we would want renderTrigger to optional

src/ContextMenu.tsx
```diff
@@ -107,19 +126,23 @@ interface IContextMenuItemProps {
     data: Object,
     target: HTMLElement
   ) => void | Function;
+  menuItems?: IContextMenuItem[];
+  attributes?: HTMLAttributes<HTMLDivElement>;
```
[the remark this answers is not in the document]
casserni: Those attributes are the ones defined here, I am not entirely sure this prop is actually needed at this moment

https://github.com/vkbansal/react-contextmenu/blob/master/docs/api.md#menuitem-

src/ContextMenu.tsx
P0lip: I am afraid we may have no reasonable solution for the given case.
We can either not pass ref at all (if one need ref, ContextMenuTrigger can be embedded manually) or simply handle that null in renderTrigger.
Both are some intermediate solution to the problem, so maybe let's leave it as it?
Just make sure the typings are correct.

src/ContextMenu.tsx
```diff
@@ -23,24 +23,38 @@ import { Box, Break, IBox, Text, useTheme } from './';
  */
 
 interface IContextMenuProps {
-  renderTrigger?: (props?: IContextMenuProps) => ReactNode | string;
+  renderTrigger: (
+    ref: React.MutableRefObject<{
+      handleContextClick: (e: MouseEvent<HTMLDivElement>) => void;
```
P0lip: Could be
```
handleContextClick: MouseEventHandler<HTMLDivElement>
```
and so on

src/ContextMenu.tsx
casserni: I say we leave it. Rendering ContextMenuTrigger separately is kind of messy. Will update the ref type to be possibly undefined so that its clearer to the end user

src/ContextMenu.tsx
casserni: Needs to stay as it :(

PR #54, opened 2019-02-04
SL-1470: Implement basic tabs in UIKit

XVincentX:
This PR implements a basic Tabs that'll be used to complete the request maker component.

Given this is nothing more than a styled wrapper for another library, I'm not completely sure what I should be testing here, so any suggestion here is welcomed!


package.json
```diff
@@ -54,6 +54,7 @@
     "react-input-autosize": "2.2.x",
     "react-select": "2.1.x",
     "react-simple-code-editor": "0.8.x",
+    "react-tabs": "^3.0.0",
```
P0lip: Let's define an exact version here (without `^`)

src/Tabs/Tab.tsx
```diff
@@ -0,0 +1,72 @@
+import { ClassNames, css } from '@emotion/core';
```
P0lip: You are missing `@jsx` pragma.

P0lip:
```suggestion
/* @jsx jsx */
import { ClassNames, css, jsx } from '@emotion/core';
```

src/Tabs/TabPanel.tsx
```diff
@@ -0,0 +1,25 @@
+import { ClassNames } from '@emotion/core';
```
P0lip:
```suggestion
/* @jsx jsx */
import { ClassNames, jsx } from '@emotion/core';
```

src/Tabs/Tab.tsx
```diff
@@ -0,0 +1,72 @@
...
+        </Tab>
+      )}
+    </ClassNames>
+  );
+};
+
+StyledTab.tabsRole = 'Tab';
+
+const tabStyles = (tabTheme: { fg: string; bg: string }) => {
+  return {
```
P0lip: Could we use theme hook here?
I know it doesn't really matter, but we had such a convention in all components.

src/Tabs/Tab.tsx
```diff
@@ -0,0 +1,72 @@
...
+    tabStyle: css({
```
P0lip: You don't need `css` here.
A plain object literal is more than enough :)

package.json
```diff
@@ -68,21 +69,22 @@
...
     "@types/react-window": "1.1.x",
+    "@types/styled-system": "3.0.x",
     "react": "^16.7.0-alpha.2",
     "react-dom": "^16.7.0-alpha.2",
     "storybook-chromatic": "1.2.x",
     "typescript": "3.2.2"
   },
   "lint-staged": {
-    "xxx": [
+    "*.{ts,tsx}": [
```
P0lip: :+1:

src/Tabs/Tab.tsx
```diff
@@ -0,0 +1,72 @@
...
+      cursor: 'pointer',
+      backgroundColor: tabTheme.bg,
+      color: tabTheme.fg,
+
+      '&:focus': {
+        boxShadow: '0 0 5px hsl(208, 99%, 50%)',
+        borderColor: 'hsl(208, 99%, 50%)',
+        outline: 'none',
+
+        '&:after': {
```
P0lip: `::after` is [preferred](https://developer.mozilla.org/en-US/docs/Web/CSS/::after) over `:after`

src/Tabs/Tab.tsx
```diff
@@ -0,0 +1,72 @@
...
+      position: 'relative',
+      listStyle: 'none',
+      padding: '6px 12px',
+      cursor: 'pointer',
+      backgroundColor: tabTheme.bg,
+      color: tabTheme.fg,
+
+      '&:focus': {
+        boxShadow: '0 0 5px hsl(208, 99%, 50%)',
+        borderColor: 'hsl(208, 99%, 50%)',
```
P0lip: We are missing a themed color here.
image

You can switch between themes using our theme addon (click on theme tab)
image

src/Tabs/TabList.tsx
```diff
@@ -0,0 +1,21 @@
+/* @jsx jsx */
+
+import { css, jsx } from '@emotion/core';
+import * as React from 'react';
+import { TabList } from 'react-tabs';
```
P0lip:
```suggestion
import { TabList, TabListProps } from 'react-tabs';
```

src/Tabs/TabList.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+const StyledTabList: React.FunctionComponent & { tabsRole: string } = props => (
```
P0lip: `react-tabs` seems to export props, so let's make use of them.
```suggestion
const StyledTabList: React.FunctionComponent<TabListProps> & { tabsRole: string } = props => (
```

src/Tabs/Tab.tsx
```diff
@@ -0,0 +1,72 @@
...
+        outline: 'none',
+
+        '&:after': {
+          content: '""',
+          position: 'absolute',
+          height: '5px',
+          left: '-4px',
+          right: '-4px',
+          bottom: '-5px',
+          background: '#fff',
```
P0lip: Theme

src/Tabs/Tab.tsx
```diff
@@ -0,0 +1,72 @@
...
+        },
+      },
+    }),
+
+    selectedTabStyle: css({
+      background: '#fff',
```
P0lip: Ditto

src/Tabs/Tab.tsx
```diff
@@ -0,0 +1,72 @@
...
+      borderColor: '#aaa',
```
P0lip: Ditto

src/Tabs/Tab.tsx
```diff
@@ -0,0 +1,72 @@
...
+      color: 'black',
+      borderRadius: '5px 5px 0 0',
+    }),
+
+    disabledTabStyle: css({
+      color: 'GrayText',
```
P0lip: Ditto

src/Tabs/TabList.tsx
```diff
@@ -0,0 +1,21 @@
...
+  <TabList {...props} css={tabListStyle}>
+    {props.children}
+  </TabList>
+);
+
+StyledTabList.tabsRole = 'TabList';
+
+const tabListStyle = css({
+  borderBottom: '1px solid #aaa',
```
P0lip: Ditto

src/Tabs/Tab.tsx
P0lip: If TS complains, make sure to annotate return.
```tsx
const tabStyles = (tabTheme: { fg: string; bg: string }): Dictionary<IBoxCSS> => {
```
should do the trick for you

src/Tabs/TabPanel.tsx
XVincentX: Is this required even though I'm not using any special feature of emotion's `jsx?`

src/Tabs/Tab.tsx
XVincentX: Conventions are important! I'll fix!

src/Tabs/TabPanel.tsx
P0lip: In general, it's not required in your particular case right now, but I'd add it, as we do that in the majority of our components.
Anyway, I won't complain if you leave it as is.
I believe emotion's jsx is needed only if you do the following `<span css={styles}>a</span>`

src/Tabs/Tab.tsx
XVincentX: Ah nice, I didn't know it!

src/Tabs/Tab.tsx
XVincentX: Done in 8bdac75

src/Tabs/Tab.tsx
XVincentX: Done!

XVincentX:
@P0lip Can you have another look? Hopefully I haven't forgot anything.

src/Tabs/Tab.tsx
P0lip: @XVincentX bump ^

src/Tabs/Tab.tsx
```diff
@@ -0,0 +1,76 @@
...
+      bottom: '-1px',
+      position: 'relative',
+      listStyle: 'none',
+      padding: '6px 12px',
+      cursor: 'pointer',
+      backgroundColor: theme.tabs.bg,
+      color: theme.tabs.fg,
+
+      '&:focus': {
+        boxShadow: '0 0 5px hsl(208, 99%, 50%)',
```
P0lip: let's add a theme color for this shadow as well (follow `blockQuote` naming)

src/theme/light.ts
```diff
@@ -116,6 +116,13 @@ export const lightTheme: ITheme = {
     bg: '#fff',
   },
 
+  tabs: {
+    fg: '#222',
+    bg: '#fff',
+    border: '#aaa',
+    disabledColor: 'GrayText',
```
P0lip: `disabledFg`

XVincentX:
I admit I have no taste at all for styling, that's why I was asking for a style guide yesterday :)

src/Tabs/Tab.tsx
P0lip: Apparently, it's some blue color which looks fine for both light and dark theme, so let's leave it for now.

PR #55, opened 2019-02-06
[SL-1552] Component styling

casserni:
screen shot 2019-02-05 at 6 07 04 pm

screen shot 2019-02-06 at 9 54 15 am

So, I am currently building the explorer, which will utilize ui-kit AND be in platform. That means our components need to look nice. This PR focuses mainly on the form components. It adds padding and sizing so that our components are consistent. I also changed the colors for the light theme in a few places and extracted some constants. There is still a lot of repetition in the styling for similar components and we will want to think of a way to handle that (padding, font sizing, heights,....). This was a quick cleanup so we can start using the components in production and will go back and refactor at some point.

I also removed the convention of
```
IComponent extends IComponentProps {}
```
in theory this is a nice extraction but we aren't using it in any meaningful way and just adds extra complexity/code

IMPORTANT CHANGES
for record keeping purposes: The agreed solution by @P0lip and I is to rename defaultCSS to css then remove the emotion/jsx pragma where not completely necessary. We would also like to reduce the use of the pragma in all of our applications.

The purpose of removing the pragma is two fold
- the pragma/jsx creates a classname based on the `css` prop. Since box is a forward ref that className gets applied rather than the one created in box!! With these update box will now account for the css prop during its className generation
- secondly it will force our engineers to use ui-kit more closely to how its designed (this will also help  us identify any friction points and fix them)


marbemac:
Will leave others to review, but the screenshot looks nice! Only suggestion I have on the style in the screenshot is:

- In select: equal padding on left/right side of dropdown caret

marbemac:
A good start btw. In general I would encourage everybody to start polishing ui-kit components (and usage of these components) as they work on their stories. This will let us iteratively start improving the UI in studio, etc, in advance of a dedicated push to polish.

src/Button.tsx
```diff
@@ -6,53 +6,57 @@ import { FunctionComponent } from 'react';
...
+export const buttonStyles = ({ disabled }: IButton = {}): IBoxCSS => {
+  const { button } = useTheme();
 
   return [
     {
+      color: button.fg,
+      backgroundColor: button.bg,
+      borderColor: button.border,
+
+      padding: '5px 10px',
```
P0lip: This is a breaking change.
I used the button component on different occasions. Plenty of my buttons were very small, i.e they had only an icon inside.

src/Checkbox.tsx
```diff
@@ -16,23 +21,21 @@ export const Checkbox: FunctionComponent<ICheckbox> = props => {
...
     setValue(target.checked);
-    if (onChange !== undefined) {
-      onChange(target.checked);
-    }
+    if (onChange) onChange(target.checked);
   }, []);
 
   return (
-    <Flex {...rest} as="label" css={css} htmlFor={id}>
+    // @ts-ignore FIXME border-box
```
P0lip: To be fixed

src/Icon.tsx
```diff
@@ -43,3 +33,11 @@ export interface IIconProps {
   transform?: string | Transform;
   symbol?: FaSymbol;
 }
+
+export const Icon: FunctionComponent<IIcon> = props => {
+  return jsx(Box, {
```
P0lip: This jsx fn call is some leftover, let's use proper jsx syntax here :)

P0lip:
> There is still a lot of repetition in the styling for similar components and we will want to think of a way to handle that (padding, font sizing, heights,....).

Yeah, I agree, in particular, Toggle and Checkbox are fairly similar. Same applies to Textarea and Input.

src/Button.tsx
casserni: Updated the above image to include a button with an icon. Let me know what you think

P0lip: It looks very nice, but I am afraid it's just too big :(
Could we introduce `size` prop or something similar?

casserni:
@P0lip for the `XYZProps` I am cool with adding them on an on need basis rather than making it a convention. Can you point out which are needed currently

src/Button.tsx
casserni: @P0lip I think having a size prop for all of our components would be ideal and something we should for sure add at some point. Can you show me or include an image where this would currently be breaking? I would like to get this repo and pr moving so we can start use the components practically. 

Unfortunately since we aren't using the components meaningfully (right now they are just slapped in as replacements in studio) its hard to tell what needs to be improved upon.

P0lip: You can take a look at Slate's Toolbar, for instance. It has fairly small buttons.

P0lip:
@casserni
There was a case in Image component, but I didn't really use them that often.
Happy to stick to the convention without XYZProp.

src/Box.tsx
```diff
@@ -65,17 +67,22 @@ export const Box = forwardRef<HTMLOrSVGElement, IBox<HTMLOrSVGElement>>((props,
     textTransform,
     color,
     backgroundColor,
+    borderColor,
     transform,
     visibility,
+
+    defaultCSS,
```
P0lip: This is breaking :/

Some 'non UI-Kit' components pass css directly.

casserni: Those components should be updated to either use the `style` prop or the `defaultCSS` prop. There is too much "magic" involved with the css prop for me to be comfortable using it in such a broad scope especially in non-ui kit components. If the css prop is used there is the issue described above where prop values won't get applied  and that can cause a lot of confusion

casserni: Main difference between `defaultCSS` is overridable where as `style` is not. So its up to the end user to decide which to apply. But `css` should no longer be used

P0lip: I know there is some Emotion magic behind it, but changing it in all projects will be painful unless we have come up with some codemod for that.
I could write such a codemod if needed (a matter of few mins), but in general I'd like to involve more people to the discussion as IMHO this is quite a significant change and people might be surprised css prop doesn't work anymore.

@marbemac your thoughts?

casserni: The css prop should still work, we never had logic in Box that handles it. That was handled by the jsx/emotion pragma. The typing was removed so that I guess would be breaking since projects won't compile. So how about we add that back, leave `defaultCSS`, remove `style` (since functionally it is doing the same as `css`)

casserni: Then once we create better and more organized documentation we make sure to inform everyone the `css` would like html `style` and will override and props? @P0lip

P0lip: yeah, that sounds fine to me Nick

marbemac: I skimmed over this - sounds like ya'll figured it out! Ping me again if a third mind is needed :).

marbemac:
Where are we at on this one?

casserni:
@marbemac Unless @P0lip has anymore comments I think we should be good to merge

marbemac:
Alright, leaving to @P0lip to give last pass if needed and then approve!

PR #60, opened 2019-02-14
feat: add invalid variants for Checkbox, Input, Select, and Textarea [SL-1788]

wmhilton:
Add an optional `invalid` prop to `Checkbox`, `Input`, `Select`, and `Textarea` that activates the invalid style variant.

Normal, Light:
image

Invalid, Light:
image

Normal, Dark:
image

Invalid, Dark:
image

### DIFF
--- a/src/Select.tsx
+++ b/src/Select.tsx
@@ -133,22 +133,23 @@ export const Select: React.FunctionComponent<ISelect> = props => {
  */
 
 const customStyles = (invalid: boolean) => {
-  const { select: selectTheme } = useTheme();
-
-  if (!selectTheme) {
+  const { select: baseTheme } = useTheme();
+  if (!baseTheme) {
     return {};
   }
+  const { chip: chipTheme, menu: menuTheme, invalid: invalidTheme = {} } = baseTheme;
 
-  const { chip: chipTheme, menu: menuTheme } = selectTheme;
+  const theme = { ...baseTheme };
+  if (invalid) Object.assign(theme, invalidTheme);
 
   return {
     clearIndicator: (provided: any, { isDisabled }: { isDisabled: boolean }) => ({
       ...provided,
-      color: selectTheme.border || selectTheme.fg,
+      color: theme.border || theme.fg,
       padding: '0px',
 
       ':hover': {
-        color: selectTheme.border || selectTheme.fg,
+        color: theme.border || theme.fg,
         opacity: !isDisabled && 0.6,
       },
     }),
@@ -158,13 +159,9 @@ const customStyles = (invalid: boolean) => {
     }),
     control: (provided: any, { isDisabled }: { isDisabled: boolean }) => ({
       ...provided,
-      color: selectTheme.fg,
-      backgroundColor: selectTheme.bg,
-      border: invalid
-        ? `1px solid ${selectTheme.invalidFg}`
-        : selectTheme.border
-          ? `1px solid ${selectTheme.border}`
-          : 'none',
+      color: theme.fg,
+      backgroundColor: theme.bg,
+      border: theme.border ? `1px solid ${theme.border}` : 'none',
 
       minWidth: '147px',
       minHeight: '30px',
@@ -180,29 +177,29 @@ const customStyles = (invalid: boolean) => {
       opacity: isDisabled && 0.6,
 
       ':hover': {
-        borderColor: invalid ? selectTheme.invalidFg : selectTheme.border,
+        borderColor: theme.border,
       },
     }),
     dropdownIndicator: (provided: any, { isDisabled }: { isDisabled: boolean }) => ({
       ...provided,
-      color: selectTheme.border || selectTheme.fg,
+      color: theme.border || theme.fg,
 
       padding: '0px',
 
       ':hover': {
-        color: selectTheme.border || selectTheme.fg,
+        color: theme.border || theme.fg,
         opacity: !isDisabled && 0.6,
       },
     }),
     indicatorSeparator: (provided: any) => ({
       ...provided,
-      backgroundColor: selectTheme.border || selectTheme.fg,
+      backgroundColor: theme.border || theme.fg,
       marginLeft: '5px',
       marginRight: '5px',
     }),
     input: (provided: any, { isDisabled }: { isDisabled: false }) => ({
       ...provided,
-      color: selectTheme.fg,
+      color: theme.fg,
 
       padding: '0px',
       cursor: isDisabled ? 'not-allowed' : 'pointer',
@@ -210,7 +207,7 @@ const customStyles = (invalid: boolean) => {
     }),
     loadingIndicator: (provided: any) => ({
       ...provided,
-      color: selectTheme.border || selectTheme.fg,
+      color: theme.border || theme.fg,
       padding: '0px',
     }),
     loadingMessage: (provided: any) => ({
@@ -291,13 +288,13 @@ const customStyles = (invalid: boolean) => {
     }),
     placeholder: (provided: any) => ({
       ...provided,
-      color: selectTheme.fg,
+      color: theme.fg,
 
       opacity: 0.6,
     }),
     singleValue: (provided: any) => ({
       ...provided,
-      color: selectTheme.fg,
+      color: theme.fg,
       padding: '0px',
     }),
 
--- a/src/Select.tsx
+++ b/src/Select.tsx
@@ -35,6 +35,7 @@ export interface ISelectBaseProps {
   closeOnScroll?: boolean;
 
   allowCreate?: boolean;
+  invalid?: boolean;
 }
 
 export interface ISelectProps
@@ -82,6 +83,7 @@ export const Select: React.FunctionComponent<ISelect> = props => {
     noOptionsMessage = 'No Options',
 
     allowCreate = false,
+    invalid = false,
 
     ...selectProps
   } = props;
@@ -106,7 +108,7 @@ export const Select: React.FunctionComponent<ISelect> = props => {
     onMenuScrollToBottom: onScrollToBottom,
     ...selectProps,
     // CUSTOM STYLES
-    styles: customStyles(),
+    styles: customStyles(invalid),
   };
 
   if ('loadOptions' in props && ('onCreateOption' in props || allowCreate)) {
@@ -130,7 +132,7 @@ export const Select: React.FunctionComponent<ISelect> = props => {
  * override color related
  */
 
-const customStyles = () => {
+const customStyles = (invalid: boolean) => {
   const { select: selectTheme } = useTheme();
 
   if (!selectTheme) {
@@ -158,7 +160,11 @@ const customStyles = () => {
       ...provided,
       color: selectTheme.fg,
       backgroundColor: selectTheme.bg,
-      border: selectTheme.border ? `1px solid ${selectTheme.border}` : 'none',
+      border: invalid
+        ? `1px solid ${selectTheme.invalidFg}`
+        : selectTheme.border
+          ? `1px solid ${selectTheme.border}`
+          : 'none',
 
       minWidth: '147px',
       minHeight: '30px',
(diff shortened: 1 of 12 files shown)


P0lip:
Didn't spend much time looking into this, but you should use the following pattern for theme values naming `stateFg`, `stateBg` etc., so in your case it would be `invalidFg`. :)

src/theme/light.ts
```diff
@@ -10,13 +10,15 @@ const inputTheme: ITheme['input'] = {
   fg: '#222',
   bg: '#fff',
   border: opacityBorder,
+  invalid: 'red',
```
P0lip: `invalidFg`

Same goes for all other similar cases.

P0lip:
See comment above.

src/Select.tsx
```diff
@@ -158,7 +160,11 @@ const customStyles = () => {
       ...provided,
       color: selectTheme.fg,
       backgroundColor: selectTheme.bg,
-      border: selectTheme.border ? `1px solid ${selectTheme.border}` : 'none',
+      border: invalid
```
P0lip: Shall we also make use of [:invalid](https://developer.mozilla.org/en-US/docs/Web/CSS/:invalid) pseudo-class?
We could make use of browser validation (email, pattern etc.) and have similar styling.

src/Textarea.tsx
```diff
@@ -32,14 +33,14 @@ export const Textarea: React.FunctionComponent<ITextarea> = props => {
...
-export const textareaStyles = ({ autosize, disabled }: ITextarea): IBoxCSS => {
+export const textareaStyles = ({ autosize, disabled, invalid }: ITextarea): IBoxCSS => {
   const { textarea } = useTheme();
 
   return [
     {
       color: textarea.fg,
       backgroundColor: textarea.bg,
-      border: textarea.border ? `1px solid ${textarea.border}` : 'none',
+      border: invalid ? `1px solid ${textarea.invalidFg}` : textarea.border ? `1px solid ${textarea.border}` : 'none',
```
P0lip: not sure whether `invalidBorder` isn't a more appropriate name here.

src/Select.tsx
wmhilton: The main problem with the `:invalid` pseudo-class is it applies before elements have experienced any user interaction. Bit of a design flaw honestly, because then an empty form is bombed with tons of red.

Reading through all the options on https://stackoverflow.com/questions/7920742/delay-html5-invalid-pseudo-class-until-the-first-event it appears one *possible* solution for modernish browsers is:

```
:not(:placeholder-shown):invalid
```

But... that only works if you set a placeholder, and doesn't handle the `required` case.

My personal attitude would be: screw `:invalid` pseudo-class because it's useless, and wait for [`:user-invalid`](https://drafts.csswg.org/selectors-4/#user-pseudos) when we get CSS4.

src/Textarea.tsx
wmhilton: that sent me down a rabbit hole. anyhoo.... we got a more scalable approach now. Instead of invalid being a single color, it's a complete sub-schema:
```ts
  input: {
    fg: string;
    bg: string;
    border?: string;
    invalid?: {
      fg?: string;
      bg?: string;
      border?: string;
    };
  };
```

src/theme/dark.ts
```diff
@@ -8,8 +8,13 @@ const opacityBorder = 'rgba(16,22,26,.15)';
 
 const inputTheme: ITheme['input'] = {
   fg: '#f5f8fa',
-  bg: 'rgba(92, 92, 92, 0.5)',
+  bg: 'rgb(65, 65, 65)',
   border: opacityBorder,
+  invalid: {
```
P0lip: @casserni wanna share your concerns about this pattern?

src/Textarea.tsx
casserni: So the original thought when building the theme structure was to avoid deep nesting when possible. There are a few reasons
- First reason is that when merging deep objects its easy to accidentally overwrite and delete entire subtrees of information if you aren't careful with your merge methods. @wmhilton if you remember in the beginning you build that inject function or something to handle merging and it starts to get messy as the the theme get more complex. 

- Also along these lines with deeply nested objects we need to check the existence of parents. So select.menu.item.hover.bg? does item exist? does menu exist? does select exist?!! Then when actually using these colors you have to call the entire path instead of something like select.menuItemHoverBg

- For components made of multiple working parts (like select) having nested pseudo classes becomes confusing for what is applied when to what part. So select has chips, menus, inputs, icons. When we start nesting, each of those is their own subtree with their own pseudo class subtree and it starts to become inconsistent and again difficult to manage

Imo nesting the theme like above makes it LESS scalable. The only thing it adds is readability and organization. But organization and readability can be improved with better naming conventions and better file organization (spacing and grouping)

casserni: That being said, we have already broken our own rule of not allowing nesting. 

current:
```
  select: {
    fg: string;
    bg: string;
    border?: string;

    chip: {
      fg: string;
      bg: string;
      border?: string;
    };

    menu: {
      fg: string;
      bg: string;
      border: string;

      hoverBg: string;
      selectedBg: string;
    };
  };
```

what its intended to be:
```
  select: {
    fg: string;
    bg: string;
    border?: string;

    chipFg: string;
    chipBg: string;
    chipBorder?: string;

    menuFg: string;
    menuBg: string;
    menuBorder: string;

    menuItemHoverBg: string;
    MenuIemSelectedBg: string;
  };
```

P0lip: In general, my feelings are slightly towards the more nested approach. It certainly has some implications when you actually extend the theme, however. We'd need to provide some abstraction over it, i.e. merge the properties (same as we do in case of theme zones).
That said, the merging function would need to stay simple, otherwise we end up with:

> * if you remember in the beginning you built that inject function or something to handle merging and it starts to get messy as the the theme get more complex.

and that would be messy.


> * So select.menu.item.hover.bg? does item exist? does menu exist? does select exist?

TypeScript got us covered?


That said, I wouldn't change the current approach right now. We had quite a lengthy discussion about it some time ago and each approach has its pros and cons.
My vote would be on nested structure, but I am fine with flat either. :)

So, Will, could we stick to convention in this PR? :)

wmhilton: > So select.menu.item.hover.bg? does item exist? does menu exist? does select exist?!!

That's assuming you do `select.menu.item.invalid.hover.bg`. It really doesn't look too bad if you use it like this:

```js
const { select: { menu: { item: itemTheme } } } = theme;
const { hover: hoverTheme = {} } = itemTheme;
const { invalid: invalidTheme = {} } = itemTheme;
const { hover: invalidHoverTheme = {} } = invalidTheme;

const theme = {...item}
if (isHovering) Object.assign(theme, item.hover)
if (isInvalid) Object.assign(theme, item.invalid)
if (isHovering && isInvalid) Object.assign(theme, invalidHoverTheme)

const styles = {
  color: theme.fg,
  backgroundColor: theme.bg,
  border: theme.border ? `1px solid ${theme.border}` : 'none',
}
```

Now `theme` just contains exactly what you want: `{fg, bg, border}`

while the alternative, ends up looking like this doesn't it?:

```js
const styles = {
  color: invalid && theme.invalidFg ? theme.invalidFg : theme.fg,
  backgroundColor: invalid && theme.invalidBg ? theme.invalidBg : theme.bg,
  border: invalid
    ? theme.invalidBorder ? `1px solid ${theme.invalidBorder}` : 'none'
    : theme.border ? `1px solid ${theme.border}` : 'none',
  ':hover': {
    color: invalid && theme.invalidHoverFg ? theme.invalidHoverFg : theme.hoverFg,
    backgroundColor: invalid && theme.invalidHoverBg ? theme.invalidHoverBg : theme.hoverBg,
    border: invalid
      ? theme.invalidHoverBorder ?  ? `1px solid ${theme.invalidHoverBorder}` : 'none'
      : theme.hoverBorder ? `1px solid ${theme.hoverBorder}` : 'none',
  }
}
```

wmhilton: > So, Will, could we stick to convention in this PR? :)

Sure. I'll do that and try not to make it toooo ugly.

casserni: Not quite right with your example because we still need the the pseudo tree in the styles object, but I get the gist of it. What exactly is the argument for nesting?

wmhilton: Basically my argument is: variants like "invalid" should let you copy & paste the current style and modify it. And it's simple to override:

```js
if (invalid) Object.assign(baseTheme, variantTheme)
```

this other way, every time we create a new variant we have to define new properties for each.

wmhilton: yeah I know the examples weren't 100% correct. There's probably a way to do the second without a `:hover`... or to do the first _with_ a `:hover`... I got kinda lost lol.

wmhilton: so here's my new strategy:

```js
  const { checkbox: baseTheme } = useTheme();

  const invalidTheme = {
    fg: baseTheme.invalidFg,
    bg: baseTheme.invalidBg,
    border: baseTheme.invalidBorder,
    checked: baseTheme.invalidChecked,
  };

  const theme = { ...baseTheme };
  if (invalid) Object.assign(theme, invalidTheme);
```

Still pretty simple. 👍

casserni: I am being swayed lol So put together a button theme for example 

```
interface IButtonTheme {
  fg: string;
  bg: string;
  border?: string;

  // psudeo
  hover?: IButtonTheme;
  active?: IButtonTheme;
  focus?: IButtonTheme;

  // variants
  invalid?: IButtonTheme;
}

const ButtonTheme = {
  fg: 'black',
  bg: 'white',
  border: 'black',

  // actual pseudo classes
  hover: {
    fg: 'grey',
    bg: 'lightgrey',
    border: 'grey',
  },

  // actual varints classes
  invalid: {
    fg: 'red',
    bg: 'rose',
    border: 'red',

    // need to reset so the orginal hover doesn't override
    hover: {
      fg: 'red',
      bg: 'rose',
      border: 'red',
    },
  },
};
```

casserni: So the only thing here is we need to handle cases like the overlapping over in the theme, which I LIKE because it forces us to be more explicit and then things are more transparent :)

wmhilton: To prevent namespace conflicts between props, pseudos, and variants, maybe even:
```js
interface IButtonTheme {
  fg: string;
  bg: string;
  border?: string;

  pseudo: {
    hover?: IButtonTheme;
    active?: IButtonTheme;
    focus?: IButtonTheme;
  },
  variants: {
    invalid?: IButtonTheme;
  }
}
```

casserni: Another thought. Do you guys feel it could be helpful to update the useTheme hook to have arguments (theme path), and handle nonexistence cases?

useTheme('button.invalid') => returns the button theme with the invalid variant?
and or
useTheme(['button', 'invalid'])

wmhilton: oooooooh. Yeah. First syntax though - second one makes me think it's grabbing _both_ the "button" and the "invalid" themes.

(as problematic as `.` syntax is for working with pure user-submitted JSON... I'm 99.99% sure our themes won't have special characters we have to worry about escaping `"."` etc.)

P0lip: Not sure if it makes sense to point to actual property. However it might make sense to point to an object.

casserni: Also while we are here do you guys like the names `fg` and `bg` or are you indifferent?

wmhilton: so... hmm. Let's keep the discussion going.... somewhere? where's a good place... Jira is best but it sucks. Slack is better but still a pain for sharing code snippets.

P0lip: It's okay for me. Doesn't involve much typing and looks clean enough.

wmhilton:
Aside from the fact I accidentally triggered a discussion on redesigning the theme schema... I feel pretty good about this PR right now. It does more than just red borders: it actually changes the text to red and makes the background a little red, and the checkbox actually has a cool look as well.

imageimage

casserni:
@wmhilton yep didn't mean to hijack the thread looks good, and we can update theme formatting at some point when we flesh out our thought 👍

PR #64, opened 2019-02-19
feat: SL-1762 Badge element

StefanDywersant:
Issue: https://stoplightio.atlassian.net/browse/SL-1762

### Example
zrzut ekranu 2019-02-19 o 16 53 53

### DIFF
--- a/src/Badge.tsx
+++ b/src/Badge.tsx
@@ -1,0 +1,57 @@
+import * as React from 'react';
+
+import { Box, IBox } from './Box';
+
+export enum BadgeVariant {
+  Pill = 'pill',
+  Textual = 'textual',
+}
+
+export enum BadgeColor {
+  Error = 'error',
+  Warning = 'warning',
+  Default = 'default',
+}
+
+const colorMap = {
+  [BadgeColor.Error]: '#dc3546',
+  [BadgeColor.Warning]: '#ffc108',
+  [BadgeColor.Default]: '#6c757e',
+};
+
+export interface IBadge extends IBox<HTMLElement> {
+  label: string;
+  fixedWidth?: boolean;
+  inverse?: boolean;
+  listItem?: boolean;
+  variant?: BadgeVariant;
+  color?: BadgeColor;
+}
+
+export const Badge: React.FunctionComponent<IBadge> = props => {
+  const { label, color = BadgeColor.Error, variant = BadgeVariant.Pill, ...restProps } = props;
+
+  return (
+    <Box {...restProps} as="span" css={badgeStyles({ color, variant })}>
+      {label}
+    </Box>
+  );
+};
+
+const badgeStyles = ({ color, variant }: { color: BadgeColor; variant: BadgeVariant }) => [
+  {
+    padding: '.25em .6em',
+    fontSize: '75%',
+    fontWeight: 700,
+    lineHeight: 1,
+    borderRadius: '10rem',
+  },
+  variant === BadgeVariant.Pill && {
+    color: color === BadgeColor.Warning ? '#21252a' : 'white',
+    backgroundColor: colorMap[color],
+  },
+  variant === BadgeVariant.Textual && {
+    color: colorMap[color],
+    backgroundColor: 'transparent',
+  },
+];
--- a/src/Badge.tsx
+++ b/src/Badge.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 
-import { Box, IBox } from './Box';
+import { Box, IBox, useTheme } from './';
 
 export enum BadgeVariant {
   Pill = 'pill',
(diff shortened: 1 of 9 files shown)


src/Badge.tsx
```diff
@@ -0,0 +1,57 @@
...
+  Textual = 'textual',
+}
+
+export enum BadgeColor {
+  Error = 'error',
+  Warning = 'warning',
+  Default = 'default',
+}
+
+const colorMap = {
```
wmhilton: Can we move the color definitions into the theme?

src/Badge.tsx
```diff
@@ -0,0 +1,57 @@
...
+const badgeStyles = ({ color, variant }: { color: BadgeColor; variant: BadgeVariant }) => [
+  {
+    padding: '.25em .6em',
+    fontSize: '75%',
+    fontWeight: 700,
+    lineHeight: 1,
+    borderRadius: '10rem',
+  },
+  variant === BadgeVariant.Pill && {
+    color: color === BadgeColor.Warning ? '#21252a' : 'white',
```
wmhilton: more hardcoded colors, oops?

src/Badge.tsx
```diff
@@ -0,0 +1,57 @@
...
+  [BadgeColor.Error]: '#dc3546',
+  [BadgeColor.Warning]: '#ffc108',
+  [BadgeColor.Default]: '#6c757e',
+};
+
+export interface IBadge extends IBox<HTMLElement> {
+  label: string;
+  fixedWidth?: boolean;
+  inverse?: boolean;
+  listItem?: boolean;
```
P0lip: Doesn't seem to be used?

src/Badge.tsx
```diff
@@ -0,0 +1,57 @@
...
+  );
+};
+
+const badgeStyles = ({ color, variant }: { color: BadgeColor; variant: BadgeVariant }) => [
+  {
+    padding: '.25em .6em',
+    fontSize: '75%',
+    fontWeight: 700,
+    lineHeight: 1,
+    borderRadius: '10rem',
```
P0lip: Shouldn't be absolute value?

src/Badge.tsx
```diff
@@ -0,0 +1,57 @@
...
+const colorMap = {
+  [BadgeColor.Error]: '#dc3546',
+  [BadgeColor.Warning]: '#ffc108',
+  [BadgeColor.Default]: '#6c757e',
+};
+
+export interface IBadge extends IBox<HTMLElement> {
+  label: string;
+  fixedWidth?: boolean;
+  inverse?: boolean;
```
P0lip: Not used?

src/Badge.tsx
```diff
@@ -0,0 +1,57 @@
...
+  variant?: BadgeVariant;
+  color?: BadgeColor;
+}
+
+export const Badge: React.FunctionComponent<IBadge> = props => {
+  const { label, color = BadgeColor.Error, variant = BadgeVariant.Pill, ...restProps } = props;
+
+  return (
+    <Box {...restProps} as="span" css={badgeStyles({ color, variant })}>
+      {label}
```
P0lip: I'd say we should go with children here so that the following usage is allowed:
```jsx
<Badge>2</Badge>
```

src/__stories__/Utilities/Badge.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { Badge, BadgeColor, BadgeVariant } from '../../Badge';
+
+export const badgeKnobs = (tabName = 'Badge'): any => {
+  const props = {
+    label: text('label', '5', tabName),
+    color: select('color', [BadgeColor.Error, BadgeColor.Warning, BadgeColor.Default], BadgeColor.Default, tabName),
+    variant: select('variant', [BadgeVariant.Pill, BadgeVariant.Textual], BadgeVariant.Pill, tabName),
+  };
+
+  return omitBy(props, val => !val);
```
P0lip: `omitBy` is not needed here. It's some leftover, new code shouldn't really use it because it has no affect in most of the cases (including yours).

src/Badge.tsx
```diff
@@ -0,0 +1,57 @@
...
+  Default = 'default',
+}
+
+const colorMap = {
+  [BadgeColor.Error]: '#dc3546',
+  [BadgeColor.Warning]: '#ffc108',
+  [BadgeColor.Default]: '#6c757e',
+};
+
+export interface IBadge extends IBox<HTMLElement> {
```
P0lip:
```suggestion
export interface IBadge extends IBox<HTMLSpanElement> {
```

src/Badge.tsx
StefanDywersant: Done

src/Badge.tsx
StefanDywersant: Moved to theme

src/Badge.tsx
StefanDywersant: 👍

src/Badge.tsx
StefanDywersant: 👍

src/Badge.tsx
StefanDywersant: 👍

src/__stories__/Utilities/Badge.tsx
StefanDywersant: 👍

src/Badge.tsx
StefanDywersant: 👍

src/Badge.tsx
StefanDywersant: 👍

src/__stories__/Utilities/Badge.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+import { withKnobs } from '@storybook/addon-knobs';
+import { select } from '@storybook/addon-knobs/react';
+import { storiesOf } from '@storybook/react';
+
+import { Badge, BadgeColor, BadgeVariant } from '../../Badge';
+
+export const badgeKnobs = (tabName = 'Badge'): any => {
+  return {
+    color: select('color', [BadgeColor.Error, BadgeColor.Warning, BadgeColor.Default], BadgeColor.Default, tabName),
```
wmhilton: Instead of manually listing them out, you could use `Object.values(BadgeColor)`.

src/Badge.tsx
```diff
@@ -0,0 +1,47 @@
+import * as React from 'react';
+
+import { Box, IBox, useTheme } from './';
+
+export enum BadgeVariant {
```
marbemac: Don't think we need to go the other direction, so can prob use `const enum` both here and in `BadgeColor`.

wmhilton: No no why wouldn't you export these?

marbemac: Oh yes did not mean to say no export - `export const enum`

src/Badge.tsx
```diff
@@ -0,0 +1,47 @@
...
+const badgeStyles = ({ color, variant }: { color: BadgeColor; variant: BadgeVariant }) => {
+  const { badge: theme } = useTheme();
+
+  return [
+    {
+      padding: '.25em .6em',
+      fontSize: '75%',
+      fontWeight: 700,
+      lineHeight: 1,
+      borderRadius: '16px',
```
marbemac: 50%

StefanDywersant: 50% will make it look like egg when there is more text.

src/__stories__/Utilities/Badge.tsx
StefanDywersant: 👍

src/Badge.tsx
StefanDywersant: First of all I get this when I change it to const enum:
`TS2475: 'const' enums can only be used in property or index access expressions or the right hand side of an import declaration or export assignment or type query.`

From what I read const enum makes enum values inlined during compilation. Because of that the code that uses this library won't work - it will compile with BadgeVariant but during runtime it will be undefined.

So finally when I use `const enum` it will work inside my library, but won't work outside.

Here is an [example](https://www.typescriptlang.org/play/index.html#src=enum%20EdgeType%20%7B%0D%0A%20%20%20%20RefersTo%20%3D%200%2C%0D%0A%20%20%20%20Generic%20%3D%2099%0D%0A%7D%20%0D%0A%0D%0Aconst%20enum%20ConstEdgeType%20%7B%0D%0A%20%20%20%20RefersTo%20%3D%200%2C%0D%0A%20%20%20%20Generic%20%3D%2099%0D%0A%7D%0D%0A%0D%0Aconsole.log(EdgeType.Generic)%3B%0D%0Aconsole.log(ConstEdgeType.Generic)%3B)

@XVincentX ^ that may interest you as well.

XVincentX: If we want the cake and eat it too, we should be exporting a simple type:

```ts
type BadgeVariant = 'pill' | 'textual'
```

StefanDywersant: Not sure if there is a lot to fight for.

- I think `enum` is handy because we hide things under nice labels :)
- we can also iterate over them (e.g. in storybook) :)
- type is not handy and it's string :(
- `const enum` inlines so cannot be used in libs :(

marbemac: Ok, let's leave as is then!

PR #66, opened 2019-02-19
fix: forwardRef all the things [SL-1805]

wmhilton:
Being able to forward a ref is a requirement for being the root node of a `renderTrigger`. So anything that you might want to wrap a `<Popup>` around needs to forward a ref.


src/Image.tsx
```diff
@@ -7,11 +7,11 @@ export interface IImage extends IBox<HTMLImageElement> {
   responsive?: boolean;
 }
 
-export const Image: React.FunctionComponent<IImage> = props => {
+export const Image = React.forwardRef<HTMLOrSVGElement, IImage>((props, ref) => {
```
P0lip:
```suggestion
export const Image = React.forwardRef<HTMLImage, IImage>((props, ref) => {
```

src/Heading.tsx
```diff
@@ -7,10 +7,10 @@ export interface IHeading extends Omit<IText<HTMLHeadingElement>, 'as'> {
   as?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
 }
 
-export const Heading: React.FunctionComponent<IHeading> = props => {
+export const Heading = React.forwardRef<HTMLOrSVGElement, IHeading>((props, ref) => {
```
P0lip:
```suggestion
export const Heading = React.forwardRef<HTMLHeadingElement, IHeading>((props, ref) => {
```

src/__stories__/Forms/Input.tsx
```diff
@@ -38,6 +38,6 @@ storiesOf('Forms:Input', module)
     <Input
       {...inputKnobs()}
       value={store.get('value')}
-      onChange={event => store.set({ value: event.currentTarget.value })}
+      onChange={(event: any) => store.set({ value: event.currentTarget.value })}
```
P0lip: Why is any needed here now?

src/Toggle.tsx
```diff
@@ -8,7 +8,7 @@ export interface IToggle extends Omit<IBox<HTMLLabelElement>, 'as|onChange'> {
   onChange?: (checked: boolean) => void;
 }
 
-export const Toggle: React.FunctionComponent<IToggle> = props => {
+export const Toggle = React.forwardRef<HTMLOrSVGElement, IToggle>((props, ref) => {
```
P0lip:
```suggestion
export const Toggle = React.forwardRef<HTMLLabelElement, IToggle>((props, ref) => {
```

src/Table.tsx
```diff
@@ -12,15 +12,15 @@ export interface ITable extends IBox<HTMLTableElement> {
   isSelected?: boolean;
 }
 
-export const Table: React.FunctionComponent<ITable> = props => {
+export const Table = React.forwardRef<HTMLOrSVGElement, ITable>((props, ref) => {
```
P0lip:
```suggestion
export const Table = React.forwardRef<HTMLTableElement, ITable>((props, ref) => {
```

src/Table.tsx
```diff
@@ -44,9 +44,9 @@ export const tableStyles = ({ isSelected }: ITable): IBoxCSS => {
 
 export interface ITableRow extends IBox<HTMLTableRowElement> {}
 
-export const TableRow: React.FunctionComponent<ITableRow> = props => {
-  return <Box {...props} as="tr" css={tableRowStyles()} />;
-};
+export const TableRow = React.forwardRef<HTMLOrSVGElement, ITableRow>((props, ref) => {
```
P0lip:
```suggestion
export const TableRow = React.forwardRef<HTMLTableRowElement, ITableRow>((props, ref) => {
```

src/Table.tsx
```diff
@@ -69,10 +69,10 @@ export interface ITableCell extends Omit<IBox<HTMLTableDataCellElement>, 'as'> {
   as?: 'th' | 'td';
 }
 
-export const TableCell: React.FunctionComponent<ITableCell> = props => {
+export const TableCell = React.forwardRef<HTMLOrSVGElement, ITableCell>((props, ref) => {
```
P0lip:
```suggestion
export const TableCell = React.forwardRef<HTMLTableDataCellElement, ITableCell>((props, ref) => {
```

src/Link.tsx
```diff
@@ -5,11 +5,11 @@ import { useTheme } from './theme';
 
 export interface ILink extends IText<HTMLAnchorElement | HTMLElement> {}
 
-export const Link: React.FunctionComponent<ILink> = props => {
+export const Link = React.forwardRef<HTMLOrSVGElement, ILink>((props, ref) => {
```
P0lip:
```suggestion
export const Link = React.forwardRef<HTMLUListElement | HTMLOListElement, ILink>((props, ref) => {
```

P0lip:
Looking good in general.

src/Input.tsx
```diff
@@ -15,7 +15,7 @@ const AutosizeWrapper: React.FunctionComponent<Partial<{ className: string }>> =
   <AutosizeInput {...props} inputClassName={className} placeholderIsMinWidth />
 );
 
-export const Input: React.FunctionComponent<IInput> = props => {
+export const Input = React.forwardRef<HTMLOrSVGElement, IInput>((props, ref) => {
```
P0lip:
```suggestion
export const Input = React.forwardRef<HTMLInputElement, IInput>((props, ref) => {
```

P0lip:
What about tabs?

src/__stories__/Forms/Input.tsx
wmhilton: I don't know! And it drives me nuts! Storybook just started to complain about this once I wrapped `Input` in a `React.forwardRef`.

There were multiple cases where I had to change an `event` to an `event: any` and each error only started happening _after_ I wrapped that particular component with a `React.forwardRef` which makes me think that wrapping them subtly changes the type.

Which if true, technically means that this PR is a BREAKING CHANGE because code that currently compiles with ui-kit _won't_ if you upgrade ui-kit because it'll start complaining about implicit `any`s? I DON'T UNDERSTAND WHAT IS HAPPENING 😱 

So I'm just shrugging it off? Mostly harmless right?

src/Image.tsx
wmhilton: I noticed that `Image` was indeed hard-coded to `img` so we _could_ be more specific here. My question is do we gain some benefit from making the type more specific?

If at some point we removed the hardcoded `img` restriction (so users could specify their own `as=SomeClass`) then we'd have to change it again. And also... IDK I'm worried that tightening the type will lead to a situation where all the ui-kit components _except_ Image work.

src/Heading.tsx
wmhilton: We can't do this because of the flexible `as` prop right?

src/Image.tsx
P0lip: Why would we ever want to remove `img` restriction for an Image component? :thinking: 
The difference might be noticeable - different (additional) properties might be exposed.

P0lip: Moreover, we do extend `HTMLImageElement` already (see `IImage` interface)

wmhilton: My question is what do we _gain_ aside from potential unintended type errors? Does making the type stricter here help in any imaginable circumstance?

src/__stories__/Forms/Input.tsx
P0lip: > Which if true, technically means that this PR is a BREAKING CHANGE because code that currently compiles with ui-kit _won't_ if you upgrade ui-kit because it'll start complaining about implicit `any`s? I DON'T UNDERSTAND WHAT IS HAPPENING

Yeah, that's what I am afraid of.

Technically the return type is indeed a bit different, but don't get why it affects event handlers (and most likely other stuff).

src/Image.tsx
wmhilton: (Same question applies to all your other type suggestion changes.)

P0lip: I mean, if we plan to stick to HTMLOrSvgElement, we can basically throw the typing away :P
This kind of means anything can be rendered and each event handler will need to have type guard.

src/__stories__/Forms/Input.tsx
wmhilton: this is the kind of TS "experience" that makes me worried that restricting the types from HTMLOrSVGElement to something more specific might mean we have to do weird casting and things.

P0lip: @wmhilton 
Actually, my "TS experience"  is "spend 2387283782 hours finding a workaround for some dumb typing problem" and eventually don't gain anything, because the code is very simple regardless.

src/Image.tsx
wmhilton: Oooooh, do you think that's why `event: any` is _needed_ now? Do you think tightening the typescript definitions might make that go away, because then it can infer the type of `event`?

wmhilton: Oooh I'm super excited about that possibility, brb gotta check

P0lip: > Oooooh, do you think that's why `event: any` is _needed_ now

Nah, tried it out and still complains.

It might be useful if you have something like this:

```ts
const onClick = useCallback<MouseEventHandler<HTMLImageElement>>((e) => {
  e.target.alt
}, []);
```

src/Input.tsx
wmhilton: We can't actually do this one right? `as` is an exposed prop for `Input`.

src/Link.tsx
wmhilton: Um... were you thinking lists?

But also `as` is an exposed prop.

src/Input.tsx
P0lip: Yeah, but this is wrong - let's no expose it. Input must be input. What else could it be? Table, paragraph?

src/Link.tsx
P0lip: Yeah, lists, sorry

src/Table.tsx
wmhilton: Could work!

src/Table.tsx
wmhilton: Could work!

src/Toggle.tsx
wmhilton: Could work!

src/Heading.tsx
wmhilton: Could work!

src/Input.tsx
wmhilton: Textarea maybe? But yeah... I'm good with removing the 'as' prop here.

P0lip: We do have a separate component for Textarea. :P

src/Link.tsx
wmhilton: Here we really might want to allow something other than `<a>` though, right? Or should we remove the 'as' prop here too?

P0lip: You are right in this case, we might want to render a different tag for some reason here.
Let's leave it as is

wmhilton:
Tabs is a wrapper around `react-tabs` and without knowing how react-tabs works under-the-hood I didn't know if it made sense to pass refs.

Looking at the documentation / source it looks like we would have to use the custom `domRef` callback attribute.

wmhilton:
Buuuuuut it would be good to support tabs I guess. Tooltips on tab names might be desirable.

wmhilton:
OK well.......... `react-tabs` only exposes `domRef` on the Tab**s** component not the Tab component, and we don't even use the Tab**s** component so... it's not really possible AFAICT.

Workaround is to use `<Popup>` _inside_ the `<Tab>` component rather than wrap it.

src/__stories__/Forms/Textarea.tsx
```diff
@@ -56,6 +56,6 @@ storiesOf('Forms:Textarea', module)
     <Textarea
       {...textareaKnobs()}
       value={store.get('value')}
-      onChange={event => store.set({ value: event.currentTarget.value })}
+      onChange={(event: any) => store.set({ value: event.currentTarget.value })}
```
P0lip: Can remove `any`?

src/__stories__/Menus/Menu.tsx
```diff
@@ -93,7 +93,7 @@ storiesOf('Menus:Menu', module)
   .add('with custom renderMenuItem', () => (
     <Menu
       {...menuKnobs()}
-      renderMenuItem={item => <span>{item.title}</span>}
+      renderMenuItem={(item: any) => <span>{item.title}</span>}
```
P0lip: `any` :(

src/__stories__/Forms/Textarea.tsx
wmhilton: Nope. :-( I guess the forwardRef breaks the type inference.

src/__stories__/Menus/Menu.tsx
wmhilton: oh fine. `(item: IMenuItem)`?

PR #67, opened 2019-02-20
[SL-1814] tabs styling

P0lip:
image

with disabled tab:

image

dark:

image

with disabled tab

image


marbemac:
Great start! Looks good from a ui perspective, will leave it to others to do the code review.

src/Tabs/Tab.tsx
```diff
@@ -12,6 +12,7 @@ const StyledTab: React.FunctionComponent<TabProps> & { tabsRole: string } = prop
   return (
     <ClassNames>
       {({ css: getClassName }) => (
+        // @ts-ignore ref type is incompatible after @types/react bump
```
XVincentX: The error here seems to be on the `css` property — isn't that related to Emotion rather than React?

In any case, I'd vote to revert the type bump; last thing I want to do is to put this statement in every component in this library.

src/Tabs/Tab.tsx
```diff
@@ -32,43 +33,40 @@ const tabStyles = (): Dictionary<IBoxCSS> => {
...
       backgroundColor: theme.tabs.bg,
       color: theme.tabs.fg,
+      zIndex: 2,
 
-      '&:focus': {
-        boxShadow: '0 0 5px hsl(208, 99%, 50%)',
-        borderColor: 'hsl(208, 99%, 50%)',
-        outline: 'none',
+      ':first-of-type': {
+        borderTopLeftRadius: '3px',
```
XVincentX: Nice!

src/Tabs/TabList.tsx
```diff
@@ -1,25 +1,22 @@
...
+import { jsx } from '@emotion/core';
 import * as React from 'react';
 
-import { Dictionary } from '@stoplight/types';
 import { TabList, TabListProps } from 'react-tabs';
-import { IBoxCSS, useTheme } from '..';
 
 const StyledTabList: React.FunctionComponent<TabListProps> & { tabsRole: string } = props => (
-  <TabList {...props} css={tabListStyle()}>
+  // @ts-ignore ref type is incompatible after @types/react bump
```
XVincentX: Yeah another one — this will propagate to other components, I suggest to revert the bump

src/Tabs/TabPanel.tsx
```diff
@@ -1,26 +1,40 @@
...
-      >
-        {props.children}
-      </TabPanel>
-    )}
-  </ClassNames>
-);
+  return (
+    <ClassNames>
+      {({ css: getClassName }) => (
+        // @ts-ignore ref type is incompatible after @types/react bump
```
XVincentX: ditto

src/Tabs/Tab.tsx
P0lip: I think the error lies in LegacyRefObject, at least this is what TS complained about on my machine.

PR #69, opened 2019-02-21
feat: add Tooltip [SL-1871]

wmhilton:
You wanted a fancier-looking tooltip? You got one! 😄 

<details><summary>Looking for the old v1 SVG tooltip gif?</summary>

</details>


wmhilton:
I will add border radii upon request. Didn't feel like messing with SVG arcs unless necessary.

marbemac:
Cool! Will leave to @P0lip or @casserni to review. Quick request though - could we call it `Tooltip`? `Callout` terminology is already taken - used for blockquotes with a specific style (danger, success, etc)

marbemac:
Well, nvm, will just review now haha. Hold please.

wmhilton:
> What about a regular box, and an absolutely positioned tooltip caret? Tooltip caret would have same bg as box, same border as box for two sides (the pointy sides).

I've done this approach before. I made a tooltip component in Vue for General Electric Aviation. The issue I ran into with this approach was there's _no_ good way to add drop-shadows that don't look awful.

Here's an example from [semantic-ui](https://semantic-ui.com/modules/popup.html). Notice how the caret doesn't have a shadow.
image

You also run into problems if you: try to use a gradient background for the tooltip, try to set _opacity_ on the tooltip but not the caret, try to set opacity on the tooltip and the caret, and the tooltip border bleeds through the caret. All the while you still have aliasing issues where the caret doesn't look as good as the rest of the div.

That's why I tried the SVG approach. Basically I knew the div + positioned caret sucks so thought better the devil I haven't tried than the one I know. Plus I learned about `shape-rendering="crispEdges"`!

wmhilton:
The SVG approach quickly became a PITA trying to support top/center/bottom/left/right so I'm switching to a div approach at least to prototype it. What do you think of this look? Note that because of the 45 degree twisted square that makes the caret, we're forced to have a padding of `sz * sqrt(2)` for the tooltip content. I've made the borders white instead of invisible so you can see.

image

wmhilton:
One nice thing about the div approach is the height auto-grows as needed, so you don't need the overflow scrolling.

wmhilton:
Hah! I'm an idiot. Or a genius? Anyway, I figured out how to overcome the big ugly margin problem.
image

wmhilton:
Not too shabby...

src/Tooltip.tsx
```diff
@@ -0,0 +1,174 @@
...
+
+  const pstyles: React.CSSProperties = {};
+  const sz = 20; // length of caret's side
+  const b = 1; // border width
+
+  pstyles.width = `${sz}px`;
+  pstyles.height = `${sz}px`;
+  pstyles.border = `${b}px solid ${theme.border}`;
+  pstyles.backgroundColor = theme.bg;
+  if (posY === 'bottom') {
```
P0lip: Could we simplify this?
I think it should be possible, as I can notice a couple of assignments could be merged.

wmhilton: Some of the assignments can be merged... but I'm not positive that makes the code easier to read / reason about. Keeping them separate is almost like a separation of concerns... but I can add some helper functions, and I think I'll do that.

wmhilton: Give it another lookover now.

PR #70, opened 2019-02-21
fix: consistently pass css through [SL-1827]

marbemac:
Fixes up most of the components to correctly pass `css` through.

### DIFF
--- a/src/Link.tsx
+++ b/src/Link.tsx
@@ -1,14 +1,15 @@
 import * as React from 'react';
-import { Box, IBox } from './Box';
+import { Box, IBox, IBoxCSS } from './Box';
 import { useTheme } from './theme';
 
 export interface ILink extends IBox<HTMLAnchorElement> {}
 
 export const Link = React.forwardRef<HTMLAnchorElement, ILink>((props, ref) => {
-  return <Box {...props} as="a" ref={ref} css={linkStyles()} />;
+  const { css, ...rest } = props;
+  return <Box {...rest} as="a" ref={ref} css={linkStyles(css)} />;
 });
 
-export const linkStyles = () => {
+export const linkStyles = (css: IBoxCSS) => {
   const { link } = useTheme();
 
   return [
@@ -26,5 +27,6 @@ export const linkStyles = () => {
         color: link.visitedFg,
       },
     },
+    css,
   ];
 };
--- a/src/__tests__/Input.spec.tsx
+++ b/src/__tests__/Input.spec.tsx
@@ -3,6 +3,7 @@ import 'jest-enzyme';
 import * as React from 'react';
 import AutosizeInput from 'react-input-autosize';
 
+import { Box } from '../Box';
 import { Input } from '../Input';
 import { useTheme } from '../theme';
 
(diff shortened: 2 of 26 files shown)


src/Link.tsx
```diff
@@ -1,14 +1,15 @@
...
-import { Box, IBox } from './Box';
+import { Box, IBox, IBoxCSS } from './Box';
 import { useTheme } from './theme';
 
 export interface ILink extends IBox<HTMLAnchorElement> {}
 
 export const Link = React.forwardRef<HTMLAnchorElement, ILink>((props, ref) => {
-  return <Box {...props} as="a" ref={ref} css={linkStyles()} />;
+  const { css, ...rest } = props;
+  return <Box {...rest} as="a" ref={ref} css={linkStyles(css)} />;
```
casserni: Minor nitpick, I noticed in most places you use css as an array, and in a few you pass css into the style function for the component, like here. Can you make this an array since thats the majority convention and css doesn't need to get used in the linkStyles function

marbemac: The `linkStyles` function actually returns an array already. The reason it's kind of mixed up is because I did the minimal amount of changes to the existing `xyzStyles` functions in the components, and some of them return objects while some return arrays. Does that make sense?

src/__tests__/Input.spec.tsx
```diff
@@ -3,35 +3,46 @@ import 'jest-enzyme';
...
   it('passes className as inputClassName to AutosizeInput component', () => {
     const wrapper = mount(<Input autosize />);
 
     expect(wrapper.find(AutosizeInput)).toHaveProp('inputClassName', expect.stringMatching('css-'));
     wrapper.unmount();
   });
+
+  describe('styles', () => {
+    // TODO: why does this not work?
+    it.skip('provides a default styling based on theme, and passes custom css through', () => {
```
wmhilton: The reason this doesn't work is it's using `mount` instead of `shallow`. Unfortunately, using `shallow` results in an error because `React.useState` isn't mocked... and that's as far as I got. My mocking skills are pretty limited.

marbemac: Yeah that's pretty much where I got as well.

@P0lip any ideas? You're the react testing expert.

P0lip: It's not that easy to explain, but in general, it's because of what's get rendered.
In this particular case, we operate on such a tree

```
  <ForwardRef css={{...}}>
      <ForwardRef as="input" type={[undefined]} value={[undefined]} onChange={[Function: handleChange]} css={{...}}>
        <ForwardRef(render) type={[undefined]} value={[undefined]} onChange={[Function: handleChange]} css={{...}} __EMOTION_TYPE_PLEASE_DO_NOT_USE__="input">
          <input type={[undefined]} value={[undefined]} onChange={[Function: handleChange]} className="css-12dyakk" />
        </ForwardRef(render)>
      </ForwardRef>
    </ForwardRef>
```

As you can see, we don't deal with actual DOM tree here. If you take a closer look, you can see the above tree is very similar to the one displayed in react dev tools.
Enzyme's find is capable of selecting these nodes that aren't actually directly represented in DOM, hence in our case, the wrapper was the top level node (actual Input (`<Input css={{ opacity: 0.5 }} />`) component from your test) and no the actual input node that would get rendered in DOM. The most nested node in the tree is the actual input node that is eventually displayed to the user. The other ones are Box and so on, basically it represents all components involved in the rendering process (yet again - this is not equal to nodes in DOM).

I know it's quite annoying & troublesome to reason about, hence I am happy to make it more obvious and bullet-proof.
Since we don't really deal with any advanced views, don't need to support react native and our components are quite simple, we can easily switch to [react-dom/test-utils](https://reactjs.org/docs/test-utils.html) (or some wrapper around it, i.e. react-testing-library).
This would let us deal with a much simpler structure and let us write tests easier.

The truth is we don't even need to mock anything, because most of components have no logic at all.

Enzyme is great and powerful (its rich feature set makes everything complicated, though), but I won't be stubborn and keep saying it's the best fit for UI-Kit. 
We might be happier with react-dom/test-utils.

PR #74, opened 2019-02-23
fix: hook usage, deps upgrade, better tslint usage

marbemac:
This PR addresses several things:

1. There were a number of places where we conditionally used hooks (see changes to `Dialog.tsx` or `ContextMenu.tsx` for a couple of example).
2. Added ~`tslint-react` and~ `tslint-react-hooks`.
3. Remove chromatic since we don't use it.
4. Upgrade some of our deps
5. Explicitly depend on jest, tslint-config-stoplight, and some other dev deps. Less indirection, more clarity.
6. Add a `build` step to our circle ci file, so that we can check that both `test` and `build` processes work before merging PRs.

~@wmhilton, do you think you could work on fixing the issues discovered by the updated linting config when you have time? I'm completely swamped. Need to evaluate which rules we wan to actually enforce, so if there is a rule that doesn't seem worth doing just post a note so that we can all discuss. Thank you!~

No huge rush on this btw, your stories for this sprint are more important.

After we've solidified the rules we want to enforce and this PR is merged we should add this linting stuff to all of our other repos that use ts and react (formtron, studio, json schema viewer, etc).

**sl-scripts thought**

Starting that it might make sense to focus the role of `sl-scripts` on: 

- bootstrap new typescript/react libs/projects 
- provide build related commands
- provide release related commands
- provide base/common configs for ts, jest, and semantic-release


src/Badge.tsx
```diff
@@ -21,12 +21,20 @@ export interface IBadge extends IBox<HTMLSpanElement> {
 export const Badge: React.FunctionComponent<IBadge> = props => {
   const { color = BadgeColor.Default, variant = BadgeVariant.Pill, ...restProps } = props;
 
-  return <Box {...restProps} as="span" css={badgeStyles({ color, variant })} />;
-};
-
-const badgeStyles = ({ color, variant }: { color: BadgeColor; variant: BadgeVariant }) => {
   const { badge: theme } = useTheme();
```
P0lip: Any reason for moving `useTheme` to the body of component?

src/BlockQuote.tsx
```diff
@@ -1,23 +1,23 @@
...
 export interface IBlockQuote extends IBox<HTMLQuoteElement | HTMLElement> {}
 
 export const BlockQuote = React.forwardRef<HTMLQuoteElement | HTMLElement, IBlockQuote>((props, ref) => {
   const { as = 'blockquote', css, ...rest } = props;
 
-  return <Box {...rest} as={as} ref={ref} css={[blockQuoteStyles(), css]} />;
-});
-
-export const blockQuoteStyles = () => {
   const { blockQuote } = useTheme();
```
P0lip: Ditto

src/ContextMenu.tsx
```diff
@@ -69,28 +69,28 @@ export interface IContextMenuView {
 export const ContextMenuView: React.FunctionComponent<IContextMenuView> = props => {
   const { menuItems = [], ...viewProps } = props;
 
+  const { contextMenu: theme } = useTheme();
```
P0lip: :+1: this is actually correct! (fewer hooks could be rendered if we return early).

P0lip:
Re rules, I'd propose the following ones:

* jsx-boolean-value
* jsx-key
* jsx-no-string-ref

and... 
* react/jsx-filename-extension
* react/jsx-pascal-case

but unfortunately, the above ones are not implemented in tslint-react-plugin (that's btw quite the issue with all these plugins aiming to bring/port ESLint counterpart... they usually don't support a bunch of rules or/and don't follow the changes made to them) 

What about accessibility plugins such as (tslint-react-a11y)[https://www.npmjs.com/package/tslint-react-a11y]? Unfortunately, it doesn't seem to be actively maintained, but still, there are a couple of annoying (but bringing value) rules.

Note, I omitted a couple of styling rules that are likely to be enforced by Prettier.

Shall we switch to ESLint? TSLint is likely to be abandoned anyway (at least according to recent news) and there are plenty of rules that are implemented only for ESLint.

src/Badge.tsx
marbemac: Yup! Using hooks in functions that are not either A) a react component or B) another hook is considered an anti-pattern (as far as I can tell). This makes sense, as it is all too easy to mistakenly invoke a hook conditionally if it is not easy to tell where it is being used. We made this mistake several times (see blockquote and context menu for a couple of examples). By sticking to using hooks in those two spots, it will be more obvious if it is being used conditionally.

The `tslint-react-hooks` ruleset I added checks this for us.

P0lip: > This makes sense, as it is all too easy to mistakenly invoke a hook conditionally if it is not easy to tell where it is being used.

Ok, you convinced me :)

Usage of `useTheme` in blockQuote was actually alright, but agree about ContextMenu and Menu. 

Actually, I was trying to point that conditional usage a couple of times, but it's still difficult to spot in all places, especially when only a part of the code (diff) is displayed, so it's great to have a linter rule for that. Also, verifying that in the app is not that trivial either, because you usually need to meet some condition / get to some state.
I committed that mistake once or twice as well, so yet again very happy to have a linter do the job for us.

marbemac: Ah, I meant `Dialog.jsx` - original pr description updated! (I discovered this issue via Studio bug).

marbemac:
> Shall we switch to ESLint? TSLint is likely to be abandoned anyway (at least according to recent news) and there are plenty of rules that are implemented only for ESLint.

Yup, seems to be an interesting option now that they've spent more time on their Typescript support! See the related PR and comment here: https://github.com/stoplightio/tslint-config-stoplight/pull/4

> jsx-boolean-value

I actually don't really like this rule, forcing `isPrimary={true}` instead of `isPrimary` is kind of annoying.

src/Badge.tsx
P0lip: Really? I believe I did place it before the `(show)` condition and tested it 'live' a couple of time in Storybook (show on/off).

P0lip:
@marbemac 
It's meant to do the opposite (disallows `foo={true}` and enforces `foo`)

marbemac:
Since this fixes a number of errors, suggest we actually do the linting stuff in separate PR. Will amend and push a merge candidate in a sec.

PR #79, opened 2019-03-01
feat: forwardRefs ScrollList [SL-1939]

P0lip:
https://github.com/stoplightio/tree-list/pull/47

### DIFF
--- a/src/ScrollList.tsx
+++ b/src/ScrollList.tsx
@@ -1,9 +1,10 @@
 /* @jsx jsx */
-import { css, jsx } from '@emotion/core';
+import { css as cssx, jsx } from '@emotion/core';
 import { forwardRef } from 'react';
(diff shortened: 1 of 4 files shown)


src/ScrollList.tsx
```diff
@@ -1,47 +1,60 @@
...
   const { width, height, ...rest } = props;
 
   return (
     <AutoSizer width={width} height={height}>
       {({ width: listWidth, height: listHeight }) => (
-        <WindowFixedSizeList {...rest} css={styles} height={listHeight} width={listWidth} />
+        <ReactWindow.FixedSizeList
+          {...rest}
+          ref={ref}
+          css={scrollListStyles(theme)}
```
marbemac: Suggest allow consumer to pass css prop to affect this (similar to approach in other components).

PR #80, opened 2019-03-01
feat: autoFocus [SL-1893]

P0lip:


src/hooks/useAutoFocus.ts
```diff
@@ -0,0 +1,24 @@
+import { MutableRefObject, Ref, useEffect, useRef } from 'react';
+
+export const useAutoFocus = <T extends Partial<{ focus: () => any }>>(autoFocus?: boolean, ref?: Ref<T>) => {
```
P0lip: I don't know whether we will need it in the end.
Might be useful if we want to persist autoFocus between re-renders, but not sure.

PR #84, opened 2019-03-07
fix: button should accept hoverFg

P0lip:


src/Button.tsx
```diff
@@ -33,6 +33,7 @@ export const buttonStyles = (theme: ITheme['button'], { disabled }: IButton = {}
 
       ':hover': {
         backgroundColor: theme.hoverBg,
+        ...(theme.hoverFg && { color: theme.hoverFg }),
```
casserni: why need the `...` while background doesnt?

P0lip: It's optional

P0lip: Could also be
```color: theme.hoverFg || themefg```

marbemac: will css not figure it out if it gets an undefined value?

P0lip: @marbemac, it does, nice.
Code updated.

P0lip:
> Looks good. However it would be nice to understand when a property is mandatory or it's optional

True! Right now we tend to mix it - it's all about common sense at the moment (unlike background you usually don't change hoverFg for buttons for hove state etc.), but it might be good to have some enforced rules.

PR #85, opened 2019-03-07
SL-1877 Get rid of unknown components

XVincentX:
- Make sure the components are exported in a separated statements
- Make sure all the forwardRef components have a named function instead of an arrow one
- Make sure all the forwardRef components have the `displayName` property set

Hopefully I didn't forget components along the way :trollface:


docs/components.md
```diff
@@ -80,3 +83,38 @@ or some custom style, i.e.
...
+#### Don't
+
+```tsx
+export const BlockQuote: React.FunctionComponent<IBlockQuote> = React.forwardRef<{}>((props, ref) => {{}}
+```
+
+#### Do
+
+```tsx
+const BlockQuote: React.FunctionComponent<IBlockQuote> = React.forwardRef<{}>(function BlockQuote(props, ref) {{}}
```
P0lip: We don't really need both.

XVincentX: That's right, we can probably get rid of the named function (although it's a nice fallback in case you forgot to set the `displayProperty`)

src/Badge.tsx
```diff
@@ -66,3 +66,5 @@ const badgeStyles = ({
     },
   ];
 };
+
```
wmhilton: looks like you missed a displayName?

XVincentX: It's not required in this specific case because it's a simple `FunctionComponent` and not a `forwardRef`

XVincentX:
Let's check that out @P0lip

XVincentX:
@P0lip Apparently it does not. The component still appears as a `forwardRef` — you really need to set its display name in case you do not want to see `forwardRef(ComponentName)`

I suggest to go with this as first step and then I'll take some time to understand whether we can get rid of the `forwardRef` for the majority of the components.

P0lip:
> @P0lip Apparently it does not. The component still appears as a forwardRef — you really need to set its display name in case you do not want to see forwardRef(ComponentName)

Well, they should appear as `forwardRef`. My idea was to get rid of refs and use `forwardRef` on demand if you need it.

tslint.json
```diff
@@ -1,3 +1,9 @@
 {
-  "extends": ["tslint-config-stoplight", "tslint-react-hooks"]
+  "extends": [
+    "tslint-config-stoplight",
+    "tslint-react-hooks"
+  ],
+  "rules": {
+    "no-shadowed-variable": false
```
P0lip: :trollface:

XVincentX:
Tabs do not need it, they're `FunctionComponents`

PR #92, opened 2019-03-25
feat: Context Menu Improvements SL-2244 SL-2245

XVincentX:
* Support for Shortcut through an additional property; it gets put at the end of the current menuItem
* Make sure the background color spans on the whole menu item

### DIFF
--- a/src/ContextMenu.tsx
+++ b/src/ContextMenu.tsx
@@ -205,7 +205,7 @@ export const contextMenuItemStyles = (
 ) => {
   return [
     {
-      padding: '5px 7px',
+      padding: '10px 14px',
       fontSize: '14px',
       borderRadius: '2px',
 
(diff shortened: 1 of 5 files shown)


XVincentX:
I thought the same, but basically this context menu has no designs so it's all about personal tastes here! Added!

src/ContextMenu.tsx
```diff
@@ -115,22 +114,35 @@ const menuStyles = (theme: ITheme['contextMenu'], hasMenuItems: boolean) => {
...
-  disabled?: boolean;
-  preventClose?: boolean;
-  onClick?: (
-    event: React.TouchEvent<HTMLDivElement> | React.MouseEvent<HTMLDivElement>,
-    data: Object,
-    target: HTMLElement
-  ) => void | Function;
-  menuItems?: IContextMenuItem[];
-}
+type ContextMenuTitle = { title?: string; shortcut?: never } | { title: string; shortcut?: string };
```
XVincentX: In case you're wondering what this stuff is about — is to make the type checker aware that if I supply a `shortcut`, then the `title` is mandatory, otherwise it's not. @P0lip 

We should do this more often.

P0lip: Yeah, I like `never` same as `unknown`. Used  `unknown` on a number of occasions lately.

XVincentX:
Let’s ban any from our codebase.

PR #93, opened 2019-03-25
fix(icon): drop icon size prop

casserni:
Repos checked that use ui-kit

- studio https://github.com/stoplightio/studio/pull/162
- json-schema-viewer https://github.com/stoplightio/json-schema-viewer/pull/8
- json-schema-editor https://github.com/stoplightio/json-schema-editor/pull/11
- tree-list https://github.com/stoplightio/tree-list/pull/51
- formtron (icon size not used)
- monaco (icon size not used)
- markdown-viewer (icon size not used)
- platform (icon size not used)

### DIFF
--- a/src/Icon.tsx
+++ b/src/Icon.tsx
@@ -7,21 +7,12 @@ import {
   library,
   PullProp,
   RotateProp,
-  SizeProp,
   Transform,
 } from '@fortawesome/fontawesome-svg-core';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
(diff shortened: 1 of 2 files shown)


P0lip:
Any rationale behind the dropping the support for `size`? font-size might not be the most reliable CSS rule to scale vector graphics. In any case, if it works fine across all browsers, I won't reject.

casserni:
@P0lip size uses fontSize under the hood so having both is kind've repetitious. Also supporting just fontsize allows us to be more consistent across all components

PR #94, opened 2019-03-26
Storybook cleanup

casserni:
The only new functionality is wrapping knob functions with clean knob which removes undefined, null, and empty strings


src/__stories__/Forms/Button.tsx
```diff
@@ -9,15 +9,23 @@ import { storiesOf } from '@storybook/react';
...
+    ...omit(boxKnobs(), 'opacity'),
+    disabled: boolean('disabled', false, tabName),
+  });
+};
 
 storiesOf('Forms:Button', module)
   .addDecorator(withKnobs)
 
-  .add('with defaults', () => <Button {...buttonKnobs()}>Button Text</Button>)
+  .add('aaaisabled', () => <Button>Button Text</Button>)
```
P0lip:
```suggestion
```
I believe you didn't mean to commit this.

src/__stories__/Forms/Input.tsx
```diff
@@ -22,18 +22,22 @@ const CustomInput = forwardRef<HTMLInputElement, IInput>(function CustomInput({
...
-  invalid: boolean('invalid', false, tabName),
-});
-
-export const autosizeInputKnobs = (tabName = 'Input'): IInput => ({
-  ...inputKnobs(),
-  type: select('type', AutosizeInputType, 'text', tabName),
-});
+export const inputKnobs = (tabName = 'Input'): IInput => {
+  return cleanKnobs({
+    ...omit(boxKnobs(), 'opacity'),
```
P0lip: Why do we omit `opacity`?

src/__stories__/Forms/Textarea.tsx
```diff
@@ -7,44 +7,49 @@ import { storiesOf, StoryDecorator } from '@storybook/react';
...
 
-export const textareaKnobs = (tabName = 'Textarea'): ITextarea => ({
-  ...omit(boxKnobs<HTMLTextAreaElement>(), 'opacity'),
-  autosize: boolean('autosize', false, tabName),
-  disabled: boolean('disabled', false, tabName),
-  invalid: boolean('invalid', false, tabName),
-});
+export const textareaKnobs = (tabName = 'Textarea'): ITextarea => {
+  return cleanKnobs({
+    ...omit(boxKnobs<HTMLTextAreaElement>(), 'opacity'),
```
P0lip: Ditto

src/__stories__/_utils.ts
```diff
@@ -108,3 +110,15 @@ export const InlineInputType = [
...
+
+/**
+ * CLEAN KNOBS
+ * use this function for better cleaning of the props, we want to strip out undefined/falsey value (except zero)
+ * empty props normally get passed to the storybook component which messes with default values
+ */
+
+export function cleanKnobs<T = {}>(props: any): Partial<T> {
+  return pickBy(props, (prop: any) => {
+    return prop !== undefined && prop !== null && prop !== '';
```
P0lip: Knobs might be nested (see Popup for a reference), therefore I believe we'd need to clean knobs recursively.

src/__stories__/Forms/Button.tsx
casserni: oops

src/__stories__/Forms/Input.tsx
casserni: @P0lip honestly I have no idea? It was there before, but will remove

src/__stories__/_utils.ts
casserni: changed to use this small library called cleanDeep that does recursive removal

P0lip:
@casserni knobs don't seem to have any effect at all.

casserni:
@P0lip oops fixed

package.json
```diff
@@ -52,6 +52,7 @@
     "@fortawesome/react-fontawesome": "0.1.4",
     "@stoplight/types": "3.1.x",
     "@types/styled-system": "3.2.x",
+    "clean-deep": "^3.0.2",
```
P0lip: I believe this should be a dev dependency.

PR #95, opened 2019-03-27
feat: use a portal overlay to prevent clicks SL-2246

XVincentX:


src/ContextMenu.tsx
```diff
@@ -70,13 +72,41 @@ const ContextMenuView: React.FunctionComponent<IContextMenuView> = props => {
...
 
   return (
-    <Box {...viewProps} as={ReactContextMenu} css={menuStyles(theme, menuItems.length > 0)}>
-      {menuItems.map((item, index) => {
-        return <ContextMenuItem key={index} {...item} />;
-      })}
-    </Box>
+    <>
+      {props.blockExternalClicks && isVisible
+        ? ReactDOM.createPortal(
```
P0lip: Suggest using Portal component from UI-Kit.
```suggestion
     <Portal>
```

src/ContextMenu.tsx
```diff
@@ -70,13 +72,41 @@ const ContextMenuView: React.FunctionComponent<IContextMenuView> = props => {
...
+            <Box
+              onClick={e => {
+                e.stopPropagation();
+                setIsVisible(false);
+              }}
+              width="100vw"
+              height="100vh"
+              position="absolute"
+              zIndex={10000}
```
P0lip: Shall we increase it?
I'd say we should follow the Overlay approach - 2^31 and some offset.
```suggestion
              zIndex={2 ** 31 - 3}
```

src/ContextMenu.tsx
```diff
@@ -70,13 +72,41 @@ const ContextMenuView: React.FunctionComponent<IContextMenuView> = props => {
...
+              opacity={0}
+            />,
+            document.body
+          )
+        : null}
+
+      <Box
+        {...viewProps}
+        zIndex={10001}
```
P0lip:
```suggestion
        zIndex={2 ** 31 - 2}
```

XVincentX: I've used the know constant.

P0lip: AFAIK z-index supports 32bit integers (equals `2 ** 31 - 1`), `Number.MAX_SAFE_INTEGER` goes beyond the range.

XVincentX: Interesting fact: `z-index` is an integer and…

There is no official range of valid <integer> values. Opera 12.1 supports values up to 215-1, IE up to 220-1, and other browsers even higher. During the CSS3 Values cycle there was a lot of discussion about setting a minimum range to support: the latest decision, in April 2012 during the LC phase, was [-227-1; 227-1], but other values like 224-1 and 230-1 were also proposed. However, the latest spec doesn't specify a range anymore.

Making the change anyway!

P0lip: Yeah, I looked for the range in the spec, but couldn't find any.

Looking at the following [table](https://stackoverflow.com/a/25461690) it seems like most of the vendors use 32-bit integers.

Side fact:
In any case, it's still better to use 32-bit signed integers because of the way they are represented in engines, in particular, V8.

AFAIR `Smi` (the internal representation used by V8) can contain the full range of 32-bit signed integers on 64-bit platforms and is clamped down by `1` on 32-bit platforms, but AFAIK they are going to make it consistent.
It's, in general, more optimal than `HeapNumber`.

PR #102, opened 2019-04-08
feat: improve style extensibility

tommymorgan:
1. Separate variables from blueprint.scss so other apps can include just
the variables.
2. Compile, concatenate, and copy CSS to dist directory.


package.json
```diff
@@ -25,10 +25,10 @@
     "node": ">=8.3.0"
   },
   "scripts": {
-    "build": "yarn build.styles && sl-scripts build && cp -a ./styles/. ./dist/styles",
-    "build.styles": "yarn build.bp && yarn build.tw",
-    "build.bp": "node-sass ./styles/blueprint.scss ./styles/blueprint.css --functions ./scripts/sass-inline-svg.js --importer node_modules/node-sass-package-importer/dist/cli.js",
-    "build.tw": "postcss ./styles/tailwind.css -o ./styles/index.css",
+    "build": "rm -rf build && sl-scripts build && yarn build.styles",
+    "build.styles": "yarn build.bp && yarn build.tw && mkdir -p dist && cat ./build/blueprint.css ./build/tailwind.css > ./dist/ui-kit.css && cp ./styles/_variables.scss ./dist/",
```
casserni: is there a reason to have this extra build folder?

tommymorgan: I created it to hold ephemeral files not intended for the dist or styles directories.  I think it belongs.

casserni:
Also we need `blueprint.scss` to be exposed in the dist folder in its entirety, not just the the variables.

Reason being when those variables change we need to recompile the sass at runtime

tommymorgan:
Blueprint.css is exposed in the dist directory via ui-kit.css.  I concatenated it with tailwind.css to reduce HTTP requests from the client.

casserni:
Stale pr closing

PR #107, opened 2019-05-16
feat: add scrollToItem prop to FixedSizeList

wmhilton:
# Problem Statement

I'm trying to fix this: selecting search items with the keyboard moves the selection off-screen.

This is not a problem with Blueprint's native suggest, but is with ours.

---

# Thoughts on how to fix

I think I can use the `onActiveItemChange` handler of `<Suggest>` so that the selected items stays in view, if we have a way to scroll list items into view. Which we _almost_ do because react-window supports it: https://react-window.now.sh/#/examples/list/scroll-to-item

But.... I'm running into problems with the custom scrollbar container.

Here it is with `outerElementType={CustomScrollContainer}` commented out.

I'm using a `scrollToItem` prop and watching it, but am flexible about how the API could be exposed. We just need to demonstrate it can be done first. :/

### DIFF
--- a/src/ScrollList/index.tsx
+++ b/src/ScrollList/index.tsx
@@ -35,14 +35,21 @@ export const CustomScrollContainer = React.forwardRef<HTMLDivElement, IFixedSize
 interface IFixedSizeListProps extends Omit<ReactWindow.FixedSizeListProps, 'height' | 'width'> {
   className?: string;
   maxRows?: number;
+  scrollToItem?: number;
 }
 
 const FixedSizeList: React.FunctionComponent<IFixedSizeListProps> = React.forwardRef<
   ReactWindow.FixedSizeList,
   IFixedSizeListProps
 >(function FixedSizeList(props, ref) {
-  const { className, children, itemSize, itemCount, maxRows, style, ...rest } = props;
+  const { className, children, itemSize, itemCount, maxRows, style, scrollToItem, ...rest } = props;
   const listHeight = (min([itemCount, maxRows]) as number) * itemSize;
+
+  const listRef: React.Ref<any> = React.useRef();
+
+  React.useEffect(() => {
+    if (listRef.current && typeof scrollToItem === 'number') listRef.current.scrollToItem(scrollToItem);
+  }, [listRef, scrollToItem]);
 
   return (
     <div style={{ height: maxRows ? listHeight : '100%' }} className="ScrollList-Container">
@@ -59,7 +66,8 @@ const FixedSizeList: React.FunctionComponent<IFixedSizeListProps> = React.forwar
             // style gets passed to SrollList-Scrollbars
             style={style}
             outerRef={ref}
-            outerElementType={CustomScrollContainer}
+            ref={listRef}
+            // outerElementType={CustomScrollContainer}
           >
             {children}
           </ReactWindow.FixedSizeList>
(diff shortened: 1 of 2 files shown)


src/ScrollList/index.tsx
```diff
@@ -35,15 +35,22 @@ export const CustomScrollContainer = React.forwardRef<HTMLDivElement, IFixedSize
...
 
 const FixedSizeList: React.FunctionComponent<IFixedSizeListProps> = React.forwardRef<
   ReactWindow.FixedSizeList,
   IFixedSizeListProps
 >(function FixedSizeList(props, ref) {
-  const { className, children, itemSize, itemCount, maxRows, style, ...rest } = props;
+  const { className, children, itemSize, itemCount, maxRows, style, scrollToItem, ...rest } = props;
   const listHeight = (min([itemCount, maxRows]) as number) * itemSize;
 
+  const listRef: React.Ref<any> = React.useRef();
```
P0lip: I'd ditch controlled mode and simply expose the method via forwarded ref.
Then, in any component, you could simply do the follownig:

```
const ref = React.useRef<FixedSizeList>();

<FixedSizeList listRef={ref} {...props} />

ref.current.scrollToItem(index);
```
I'm not sure, though, whether react-window executes the `onScroll` handler in such case. If it doesn't, the custom scroll container won't retrieve the dispatched scroll event and hence won't scroll.
You might need to trigger it manually
https://www.npmjs.com/package/react-scrollbars-custom#instance-methods

wmhilton:
Here's what happens if I uncomment the custom scroll container:

src/ScrollList/index.tsx
wmhilton: yeah that's fine, that's what I had originally. the prop just seemed cleaner.

> If it doesn't, the custom scroll container won't retrieve the dispatched scroll event and hence won't scroll.

so....... how would I intercept the scroll? wrap it?

wmhilton: ooooh hmm hang on that's inverted. it's calling onScroll when it scrolls. hmm

P0lip: > it looks like it's already wrapping the scroll container and adding an onScroll handler :/

Yup.
The question is - is the `onScroll` handler executed or not? I haven't verified that yet, but have you perhaps?

wmhilton: the onScroll handler is executed when the _user_ scrolls. I haven't figured out how (if at all) the FixedSizedList emits scrolling events.

I'm like in `ref` hell debugging this. I like have 4 refs now, and I don't know which ones are doing what. 😆

wmhilton: the FixedSixedList is trying to set `scrollTop` of `outerRef`

P0lip: > I haven't figured out how (if at all) the FixedSizedList emits scrolling events.

ah, sorry, I didn't mean that. What I was wondering about was whether `onChange` is executed when you call `scrollToItem` method. I assumed it's not and therefore ScrollContainer is not scrolled. You just confirmed that, so we are on the same page now.

> the FixedSixedList is trying to set scrollTop of outerRef

 so... how about wrapping `scrollToItem`?
Our own implementation would do the following:
- call the original impl of react-window's scrollToItem
- measure where we are at (scrollTop might be a good candidate here)
- call scrollToSomething method, but on scroll container this time

Thoughts?

wmhilton:
I think this was closed by mistake when the v2 branch was deleted.

P0lip:
The support for that landed in tree-list, hence Marc closed it.

PR #110, opened 2019-05-17
fix: dark tabs

lottamus:
image


src/styles/components/SimpleTabs/_base.scss
```diff
@@ -56,8 +45,71 @@
...
+
+  &.#{$ns}-dark,
+  .#{$ns}-dark & {
+    color: get-color(gray, 5);
+    background-color: #293742;
+    border-color: rgba(16, 22, 26, 0.4);
+
+    &.selected-tab {
+      color: white;
+      background-color: #30404c;
```
lottamus: @casserni what do you want to do about these colors? https://github.com/stoplightio/ui-kit/pull/110/files#diff-5d757190a8e036540b65e00b1afcd45fR34

src/styles/components/SimpleTabs/_base.scss
```diff
@@ -56,8 +45,71 @@
...
+  &.disabled-tab {
+    &:hover {
+      color: get-color(gray, 5);
+    }
+  }
+
+  &.#{$ns}-dark,
+  .#{$ns}-dark & {
+    color: get-color(gray, 5);
+    background-color: #293742;
```
lottamus: https://github.com/stoplightio/ui-kit/pull/110/files#diff-5d757190a8e036540b65e00b1afcd45fR33

src/styles/components/SimpleTabs/_base.scss
casserni: Yeah since these colors are defined as blueprint variables and not taken from the theme we need to think of something either extend the theme to have more colors or rethink our strategy

https://github.com/stoplightio/ui-kit/blob/master/src/styles/blueprint/_base.scss#L34

lottamus: Moved these colors to variables

PR #111, opened 2019-05-22
Feat/form button

casserni:
Added a Form Button that takes in a schema and data

if the data does not pass schema validation the button is disabled
if it does pass, that data is passed to the onClick

used in https://github.com/stoplightio/app-gen3/pull/38

### DIFF
--- a/src/FormButton/index.tsx
+++ b/src/FormButton/index.tsx
@@ -4,6 +4,7 @@ import * as React from 'react';
 // TODO: should probably use ajv and json schema
 import * as yup from 'yup';
 
+import { useValidateSchema } from '../_hooks/useValidateSchema';
 import { Button, IButtonProps } from '../index';
 
 /**
@@ -25,30 +26,10 @@ const FormButton: React.FunctionComponent<IFormButton> = ({
   onClick = _noop,
   ...buttonProps
 }) => {
-  const [isValid, setIsValid] = React.useState<boolean>(false);
+  const errors = useValidateSchema(schema, data);
+  const handleClick = React.useCallback(() => onClick(data), [onClick, data]);
 
-  async function validate() {
-    if (!schema) return;
-    try {
-      await schema.validateSync(data);
-    } catch (e) {
-      setIsValid(false);
-    }
-  }
-
-  React.useEffect(() => {
-    setIsValid(true);
-    validate();
-  }, [data, schema]);
-
-  return (
-    <Button
-      disabled={!isValid || loading || disabled}
-      loading={loading}
-      onClick={() => onClick(data)}
-      {...buttonProps}
-    />
-  );
+  return <Button disabled={!!errors || loading || disabled} loading={loading} onClick={handleClick} {...buttonProps} />;
 };
 
 /**
--- a/src/FormButton/index.tsx
+++ b/src/FormButton/index.tsx
@@ -1,0 +1,57 @@
+import { Omit } from '@stoplight/types';
+import _noop = require('lodash/noop');
+import * as React from 'react';
+// TODO: should probably use ajv and json schema
+import * as yup from 'yup';
+
+import { Button, IButtonProps } from '../index';
+
+/**
+ * FORM BUTTON
+ * button that is disabled if the schema doesn't pass validation
+ * TODO, add tooltip to explain why its disabled?
+ */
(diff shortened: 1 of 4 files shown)


src/FormButton/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+  React.useEffect(() => {
+    setIsValid(true);
+    validate();
+  }, [data, schema]);
+
+  return (
+    <Button
+      disabled={!isValid || loading || disabled}
+      loading={loading}
+      onClick={() => onClick(data)}
```
lottamus: Suggest pulling this out to a callback

```typescript
const handleClick = React.useCallback(() => onClick(data), [onClick, data])
```

src/FormButton/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+    try {
+      await schema.validateSync(data);
+    } catch (e) {
+      setIsValid(false);
+    }
+  }
+
+  React.useEffect(() => {
+    setIsValid(true);
+    validate();
```
lottamus: Is there a benefit in defining `validate` as a function outside of the useEffect scope? I believe defining outside will create a new function on each render and defining inside useEffect will memoize it with data and schema 🤔 

```ts
React.useEffect(async () => {
    setIsValid(true);

    if (!schema) return;

    try {
      await schema.validateSync(data);
    } catch (e) {
      setIsValid(false);
    }
}, [data, schema]);
```

src/FormInput/index.tsx
```diff
@@ -62,68 +60,26 @@ interface IFormInputValidationProps {
...
     } catch (e) {
-      setValidationError(true);
+      setValidation(e.message || 'input is invalid');
     }
   }
 
   React.useEffect(() => {
-    setValidationError(false);
+    setValidation(null);
     validate();
```
lottamus: Same as above. Just a thought 😁

src/FormButton/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+import * as React from 'react';
+// TODO: should probably use ajv and json schema
+import * as yup from 'yup';
+
+import { Button, IButtonProps } from '../index';
+
+/**
+ * FORM BUTTON
+ * button that is disabled if the schema doesn't pass validation
+ * TODO, add tooltip to explain why its disabled?
```
lottamus: Could make it optional but meh..

src/FormButton/index.tsx
casserni: useEffect, doesn't allow async functions as the input, but I can move the validate function definitions inside of the effect

lottamus: Ah yeah that makes sense because async functions always return a promise but useEffect expects a cleanup function. Yeah you could define it inside would work too!

lottamus: Or just use it as a promise lol

lottamus:
```ts
React.useEffect(() => {
    setIsValid(true);

    if (!schema) return;

    schema.validateSync(data).catch(() => setIsValid(false));
}, [data, schema]);
```

casserni: tried the catch and its not actually catching the error

src/FormButton/index.tsx
casserni: done

src/FormButton/index.tsx
casserni: Extracted to a hook

src/FormButton/index.tsx
casserni: will add when needed, I think with a combination of form inputs we can do without

src/_hooks/useValidateSchema.ts
```diff
@@ -0,0 +1,21 @@
...
+    setValidationErrors(null);
+
+    if (!schema) return;
+
+    try {
+      schema.validateSync(value, validateOpts);
+    } catch (e) {
+      setValidationErrors(e.errors || ['Input is invalid']);
+    }
+  }, [value, schema]);
```
lottamus: Probably also need to pass in `validateOpts` here as well

lottamus:
```suggestion
  }, [value, schema, validateOpts]);
```

PR #112, opened 2019-05-23
feat: SecretInput

wmhilton:
### Storybook examples:

image

image


wmhilton:
Edit: I've tweaked the typings after doing a test-integration with Studio and discovering it was lacking.

PR #114, opened 2019-05-27
feat(fixed-size-list): visual offset

P0lip:
The property might be useful if you want to add extra padding at the end of the list to highlight the fact you can scroll the list down. Didn't make it a default, as there are cases this might not be the desired behavior.

### DIFF
--- a/src/ScrollList/index.tsx
+++ b/src/ScrollList/index.tsx
@@ -35,14 +35,15 @@ export const CustomScrollContainer = React.forwardRef<HTMLDivElement, IFixedSize
 interface IFixedSizeListProps extends Omit<ReactWindow.FixedSizeListProps, 'height' | 'width'> {
   className?: string;
   maxRows?: number;
+  addOffset?: boolean;
 }
 
 const FixedSizeList: React.FunctionComponent<IFixedSizeListProps> = React.forwardRef<
   ReactWindow.FixedSizeList,
   IFixedSizeListProps
 >(function FixedSizeList(props, ref) {
-  const { className, children, itemSize, itemCount, maxRows, style, ...rest } = props;
-  const listHeight = (min([itemCount, maxRows]) as number) * itemSize;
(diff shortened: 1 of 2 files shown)


src/ScrollList/index.tsx
```diff
@@ -35,14 +35,15 @@ export const CustomScrollContainer = React.forwardRef<HTMLDivElement, IFixedSize
 interface IFixedSizeListProps extends Omit<ReactWindow.FixedSizeListProps, 'height' | 'width'> {
   className?: string;
   maxRows?: number;
+  addOffset?: boolean;
```
lottamus: I suggest changing this to `offset` so we can choose to add or remove some height from the list. It would default to `0`. Thoughts?

```suggestion
  offset?: number;
```

src/__stories__/ScrollList/index.tsx
```diff
@@ -14,6 +14,7 @@ import { FixedSizeList, IFixedSizeListProps } from '../../ScrollList';
 export const fixedSizeListKnobs = (tabName = 'FixedSizeList'): Omit<IFixedSizeListProps, 'children'> => ({
   itemCount: number('itemCount', 20, { min: 0, max: Infinity, range: false, step: 1 }, tabName),
   itemSize: number('itemSize', 50, { min: 0, max: Infinity, range: false, step: 1 }, tabName),
+  offset: number('offset', 0, { min: 0, max: Infinity, range: false, step: 1 }, tabName),
```
lottamus:
```suggestion
  offset: number('offset', 0, { min: Infinity, max: Infinity, range: false, step: 1 }, tabName),
```

lottamus: You can have a negative offset

```suggestion
  offset: number('offset', 0, { min: Infinity, max: Infinity, range: false, step: 1 }, tabName),
```

PR #115, opened 2019-05-29
feat: implement Dropdown

P0lip:


casserni:
@P0lip needs some padding or something?

Screen Shot 2019-05-29 at 11 59 52 AM

P0lip:
@casserni 
Yeah, basically it's all about having the custom scrollbar.
See https://github.com/stoplightio/studio/pull/234#discussion_r287395772

casserni:
@P0lip Gotcha, is there a way we can pass props to updated the props on the FixedSizedList? Say maybe I don't want 30 items to display or something.

P0lip:
> @P0lip Gotcha, is there a way we can pass props to change the props on the FixedSizedList? Say maybe I don't want 10 items to display or something. Or maybe I need to change the row size/height since I passed in a custom itemRenderer

Good point, I can make it customizable.

P0lip:
@casserni done! re padding - any suggestions?

casserni:
@P0lip also do we always want that minimal type styling without borders/shadows/etc on the drop down input? Also maybe a default empty message that could look nice

P0lip:
@casserni I have no clue whether we need minimal styling or not. It can be overridden if needed, but perhaps a regular variant is a better default here? Not sure.
@lottamus thoughts?

lottamus:
> I have no clue whether we need minimal styling or not. It can be overridden if needed, but perhaps a regular variant is a better default here? Not sure.

Let's keep this component simple and we can change it as needed down the road.

> needs some padding or something?

Careful with the padding since scrolling will be internal. Suggest just exposing a class name that the consumer can add their own padding

P0lip:
> Careful with the padding since scrolling will be internal. Suggest just exposing a class name that the consumer can add their own padding

Your suggestion is valid. Ttat's how I handle it in that Studio PR (`popoverProps` + `className`) BTW

PR #118, opened 2019-06-18
fix(code-viewer): syntax highlighting

P0lip:
Don't add any before/after screenshots, as there is really nothing to compare.
At the moment, everything is displayed as it was a plain text.


marbemac:
What does this fix?

It changed behavior inside of Studio (markdown code blocks are light theme now, instead of dark as they are in beta.5):

Before:

Screen Shot 2019-06-23 at 3 05 39 PM

After:

Screen Shot 2019-06-23 at 3 05 16 PM

PR #120, opened 2019-07-01
chore: add toaster story

wmhilton:
I needed to mock up some screenshots of the toast mechanism.

### DIFF
--- a/src/__stories__/Toaster/index.tsx
+++ b/src/__stories__/Toaster/index.tsx
@@ -27,6 +27,9 @@ storiesOf('Toaster', module)
     };
 
     function showToast() {
+      // I found I had to re-run these or else they wouldn't be up-to-date. 
+      // Also, Storybook doesn't let you use hooks, and frankly... I don't really know how knobs work.
+      // If there's a better way I don't know it yet.
       action = {
         icon: text('action.icon', 'airplane') as IconName,
         text: text('action.text', 'Action Text'),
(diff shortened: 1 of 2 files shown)


src/__stories__/Toaster/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+      action = {
+        icon: text('action.icon', 'airplane') as IconName,
+        text: text('action.text', 'Action Text'),
+        onClick: StorybookAction('onClick action'),
+      };
+      props = {
+        ...(boolean('show action', false) && { action }),
+        intent: select('intent', Intent, 'primary'),
+        message: text('message', 'Hello world!'),
+      };
```
wmhilton: I found I had to re-run these or else they wouldn't be up-to-date. Also, Storybook doesn't let you use hooks, and frankly... I don't really know how knobs work. There's probably a better way.

P0lip: > I found I had to re-run these or else they wouldn't be up-to-date. 

Whatcha mean by knobs being not up to date?

wmhilton: Try commenting it out and see.

P0lip: Ah yeah, I see. Hmmm, in any case, it's not a big deal. Suggest leaving a comment on that in the code though.

PR #121, opened 2019-07-03
feat: scrollbar container hover

mallachari:
Related to [SL-2672](https://stoplightio.atlassian.net/browse/SL-2672)

Changes:
- showing scrollbars when hovering over scrollable area
- fixing styling that were shadowed by default element styles
- styling consistent with vscode scrollbars

Peek 2019-07-05 09-35

### DIFF
--- a/src/ScrollContainer/index.tsx
+++ b/src/ScrollContainer/index.tsx
@@ -56,8 +56,12 @@ const ScrollContainer: React.FunctionComponent<IScrollContainer> = ({
       const shadowTopOpacity = (1 / 20) * Math.min(scrollTop, 20);
       const bottomScrollTop = scrollHeight - clientHeight;
       const shadowBottomOpacity = (1 / 20) * (bottomScrollTop - Math.max(scrollTop, bottomScrollTop - 20));
-
-      scrollbar.current.wrapperElement.style.boxShadow = `rgba(221, 221, 221, ${shadowTopOpacity}) 0px 7px 8px -7px inset, rgba(221, 221, 221, ${shadowBottomOpacity}) 0px -7px 8px -7px inset`;
+      const darkMode = window.document.getElementsByClassName('bp3-dark').length > 0;
+      scrollbar.current.wrapperElement.style.boxShadow = `rgba(${
+        darkMode ? `44, 44, 44` : `221, 221, 221`
+      }, ${shadowTopOpacity}) 0px 7px 8px -7px inset, rgba(${
+        darkMode ? `44, 44, 44` : `221, 221, 221`
+      }, ${shadowBottomOpacity}) 0px -7px 8px -7px inset`;
     },
     [scrollbar, shadows]
   );
--- a/src/ScrollContainer/index.tsx
+++ b/src/ScrollContainer/index.tsx
@@ -1,6 +1,8 @@
 import * as React from 'react';
 import { Scrollbar, ScrollbarProps } from 'react-scrollbars-custom';
 
+import omit = require('lodash/omit');
+
 import { Classes } from '../classes';
 import { AutoSizer } from '../index';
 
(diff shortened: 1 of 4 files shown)


src/ScrollContainer/index.tsx
```diff
@@ -19,25 +21,16 @@ const ScrollContainer: React.FunctionComponent<IScrollContainer> = ({
...
 
     scrollbar.current.trackXElement.style.opacity = 1;
     scrollbar.current.trackYElement.style.opacity = 1;
-
-    if (!dragEndTimeout) return;
-
-    window.clearTimeout(dragEndTimeout as number);
-    setDragEndTimeout(null);
-  }, [scrollbar, dragEndTimeout]);
+    scrollbar.current.trackXElement.style.transition = 'opacity 0s';
```
P0lip: This could be set permanently.

src/ScrollContainer/index.tsx
```diff
@@ -71,6 +61,12 @@ const ScrollContainer: React.FunctionComponent<IScrollContainer> = ({
     [scrollbar, shadows]
   );
 
+  const thumbRenderer = React.useCallback((props: any) => {
+    const { elementRef, style, className } = props;
+    const styles = omit(style, ['background', 'borderRadius']);
+    return <div className={`${className} bg-darken-5 dark:bg-darken-8`} style={styles} ref={elementRef} />;
```
P0lip: Shall we use `classnames`?

src/ScrollContainer/index.tsx
```diff
@@ -71,6 +61,12 @@ const ScrollContainer: React.FunctionComponent<IScrollContainer> = ({
     [scrollbar, shadows]
   );
 
+  const thumbRenderer = React.useCallback((props: any) => {
```
P0lip: No typings available?

src/ScrollContainer/index.tsx
```diff
@@ -1,6 +1,8 @@
 import * as React from 'react';
 import { Scrollbar, ScrollbarProps } from 'react-scrollbars-custom';
 
+import omit = require('lodash/omit');
```
P0lip:
```suggestion
import { omit } = from 'lodash';
```
We ditched the require-equals syntax.

mallachari: I saw that in other components of ui-kit and therefore used the same but agree with that.

P0lip: Yeah, it's just not updated yet.

src/ScrollContainer/index.tsx
P0lip: It'd be nice to have some transition when you hover over, so perhaps keep it but just change the value.

src/ScrollContainer/index.tsx
```diff
@@ -50,12 +43,9 @@ const ScrollContainer: React.FunctionComponent<IScrollContainer> = ({
 
     scrollbar.current.trackXElement.style.opacity = 0;
     scrollbar.current.trackYElement.style.opacity = 0;
-
-    if (!dragEndTimeout) return;
-
-    window.clearTimeout(dragEndTimeout as number);
-    setDragEndTimeout(null);
-  }, [scrollbar, dragEndTimeout]);
+    scrollbar.current.trackXElement.style.transition = 'opacity 1s';
```
P0lip: Suggest decreasing it a bit, to `0.8s` or something more or less like this.

src/ScrollContainer/index.tsx
mallachari: But then it would be permanently overridden by `opacity 1s` from `hideTracks`, wouldn't it?
Other solution could be setting classes for those two events.

P0lip: Yeah, you can keep it as is yet it would be nice to add a small transition here.

marbemac:
Looks good! Only question I have is wether we can remove the space between the scrollbar and the window edge? No worries if it's particularly difficult.

Screen Shot 2019-07-05 at 10 35 16 AM

mallachari:
> Looks good! Only question I have is wether we can remove the space between the scrollbar and the window edge? No worries if it's particularly difficult.
> 
Sure, I removed margin.

marbemac:
More more request actually - could you update the shadow style to be very dark or something when in dark mode? The light shadows look odd in dark mode.

src/styles/tailwind/_base.scss
```diff
@@ -1821,6 +1821,726 @@
   background-color: get-theme(colors , red , 5);
 }
 
+.bp3-dark .dark-hover\:bg-inherit:hover {
+  background-color: inherit;
+}
+
+.bp3-dark .dark-hover\:bg-transparent:hover {
+  background-color: get-theme(colors , transparent);
```
P0lip:
```suggestion
  background-color: get-theme(colors, transparent);
```
Could you remove the whitespace before each comma? Thank you

mallachari: Those whitespaces were generated by build process and were there originally in other classes. If I remove them, they would be overwritten within next build I suppose. Should I remove them anyway?

P0lip: Ah gotcha. No, it's okay.
Just apply the feedback from Marc regarding shadow and we are good to go!

PR #122, opened 2019-07-15
feat: expose VariableSizeList

P0lip:


src/ScrollList/index.tsx
```diff
@@ -74,31 +74,48 @@ FixedSizeList.displayName = 'FixedSizeList';
...
-// });
+  return (
+    <div className="ScrollList-Container h-full">
+      <AutoSizer>
+        {({ height, width }) => (
+          <ReactWindow.VariableSizeList
+            {...rest}
+            height={height}
+            width={width}
+            // className gets passed to ScrollList-Content
```
lottamus: These comments are a bit weird here since you're not destructing those props

P0lip: You are right. Forgot to remove them.

PR #124, opened 2019-07-19
feat: bump deps

marbemac:
Important to not get too far behind on deps. Yalc'd and tested pretty extensively inside of Studio - feels good.


src/styles/tailwind/_base.scss
```diff
@@ -9448,6 +9456,20 @@
           user-select: text;
 }
 
+.select-all {
+  -webkit-user-select: all;
```
P0lip: Don't we have autoprefixer?

marbemac: Not sure, but this file is generated during the build process - I didn't add this directly.

PR #126, opened 2019-07-30
fix: custom scrollbar container ref forwarding

mallachari:
Fixed sized list is passing reference to custom scrollbar container which needs to be forwarded further to scrollbar component. 
When it was attached to div container manipulating list with `instanceRef` was buggy.


P0lip:
LGTM, made some minor adjustments.

PR #133, opened 2019-10-08
feat: add TableOfContents component to ui kit

ssspear:
Related Issue: https://github.com/stoplightio/elements/issues/67

Screen Shot 2019-11-14 at 11 30 54 AM


src/Provider/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+const defaultHost = 'http://localhost:4060';
+export const HostContext = React.createContext(defaultHost);
+export const AxiosContext = React.createContext(axios.create());
+export const ComponentsContext = React.createContext<IComponentMapping | undefined>(undefined);
+
+const defaultIcons: NodeIconMapping = {};
+export const IconsContext = React.createContext<NodeIconMapping>(defaultIcons);
+
+export const Provider: React.FunctionComponent<IProvider> = ({ host, token, components, icons, children }) => {
```
lottamus: We can get around not moving Provider over so I think it's save to remove. We can just add support for passing a row renderer into the TableOfContents component

package.json
```diff
@@ -47,7 +47,9 @@
     "@blueprintjs/core": "^3.18.0",
     "@blueprintjs/icons": "^3.10.0",
     "@blueprintjs/select": "^3.10.0",
+    "@stoplight/markdown-viewer": "^3.7.0",
```
lottamus: We shouldn't need this since we're removing Provider

package.json
```diff
@@ -47,7 +47,9 @@
...
     "@stoplight/types": "^10.0.1",
+    "axios": "^0.19.0",
```
lottamus: Same as above

src/TableOfContents/index.tsx
```diff
@@ -0,0 +1,226 @@
+import { Button, Drawer, Icon, IconName } from '@blueprintjs/core';
+import cn from 'classnames';
+import * as React from 'react';
+import { useIsMobile } from '../_hooks/useIsMobile';
+import { ComponentsContext } from '../Provider';
+import { ScrollContainer } from '../ScrollContainer';
+import { IContentsNode } from '../types';
+import { deserializeSrn } from '../utils';
+
+export interface ITableOfContents {
```
lottamus: To get around the Provider issue above, let's add a custom row renderer so Elements can wrap each row in a custom link component:

```suggestion
export interface ITableOfContents {

  rowRenderer?: (item: IContentsNode, DefaultRow: React.FC<{ item: IContentsNode }>): React.Element;
```

In Elements the integration would look something like:

```tsx
export const TableOfContents = ({ items }) => {
  const Components = React.useContext(ComponentsContext);
  const contents = useComputeToc(items);

  return (
    <UIKitTableOfContents
      contents={contents}
      rowRenderer={(node, DefaultRow) => {
        return Components.link({ node, children: <DefaultRow item={node} /> });
      }}
    />
  );
};
```

src/TableOfContents/index.tsx
```diff
@@ -0,0 +1,226 @@
...
+  // Precomputed list of nodes instead of items. This is useful if you want to customize the ordering of the tree.
+  contents: IContentsNode[];
+
+  // SRN of the active node
+  srn?: string;
```
lottamus: I don't think we need "srn" anymore since `IContentsNode` has an `isActive` property

```suggestion
```

src/TableOfContents/index.tsx
```diff
@@ -0,0 +1,226 @@
...
+  depth: number;
+  name: string;
+  srn?: string;
+  isActive: boolean;
+  isGroup: boolean;
+  isExpanded: boolean;
+  isDivider: boolean;
+  onClick(e: React.MouseEvent<HTMLDivElement, MouseEvent>): void;
+  icon?: IconName;
+}
```
lottamus: Might be easier to just pass in the full item now that `IContentsNode` has more information on it?

```ts
interface ITableOfContentsItem {
  item: IContentsNode
  isExpanded: boolean;
  onClick(e: React.MouseEvent<HTMLDivElement, MouseEvent>): void;
}
```

src/types.ts
```diff
@@ -0,0 +1,21 @@
...
+  name: string;
+  depth: number;
+  isActive?: boolean;
+  href?: string;
+  type?: 'divider' | 'group' | 'item';
+  icon?: IconName;
+}
+
+export type IconMapType = NodeType | 'group' | 'divider' | 'item';
+export type NodeIconMapping = { [type in IconMapType]?: IconName };
```
lottamus: Can we move these types to `src/TableOfContents/types.ts`?

src/utils.ts
```diff
@@ -0,0 +1,14 @@
...
+
+  const uri = uriParts.length ? `/${uriParts.join('/')}` : '';
+
+  return {
+    service,
+    org,
+    project,
+    uri,
+  };
+}
```
lottamus: I don't think this is needed anymore. `ui-kit` doesn't need to know what an "srn" is

src/types.ts
```diff
@@ -0,0 +1,21 @@
+import { IconName } from '@blueprintjs/icons';
+import { NodeType } from '@stoplight/types';
+
+export interface IDeserializedSrn {
+  service: string;
+  org: string;
+  project: string;
+  uri: string;
+}
```
lottamus: Same here, not needed since ui-kit doesn't need to know about `srn`s

src/TableOfContents/index.tsx
```diff
@@ -0,0 +1,226 @@
...
+  );
+
+  if (isMobile) {
+    const { org, project } = deserializeSrn(srn || '');
+    return (
+      <Drawer isOpen={isOpen} onClose={() => onClose()} position="left" size="330px">
+        <div className="flex flex-1 flex-col bg-gray-1 dark:bg-transparent">
+          <div className="border-b dark:border-lighten-4 h-20 py-6 px-2 bg-white">
+            <Button className="flex justify-start text-lg" icon={'arrow-left'} minimal onClick={() => onClose()}>
+              {title || `${org} / ${project}`}
```
lottamus: We won't have "org" and "project" anymore since we're getting rid of SRN. Elements should pass in the correct title though

```
export const TableOfContents = ({ items, srn, title }) => {
  const contents = useComputeToc(items);
  const { org, project } = deserializeSrn(srn);

  return (
    <UIKitTableOfContents
      contents={contents}
      title={title || `${org} / ${project}`}
    />
  );
};
```

src/Provider/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+  return (
+    <HostContext.Provider value={host || defaultHost}>
+      <AxiosContext.Provider value={client}>
+        <ComponentsContext.Provider value={components}>
+          <IconsContext.Provider value={icons || defaultIcons}>{children}</IconsContext.Provider>
+        </ComponentsContext.Provider>
+      </AxiosContext.Provider>
+    </HostContext.Provider>
+  );
+};
```
lottamus: Can remove this file

src/TableOfContents/types.ts
```diff
@@ -0,0 +1,14 @@
...
+export interface IContentsNode {
+  name: string;
+  depth: number;
+  isActive?: boolean;
+  href?: string;
+  type?: 'divider' | 'group' | 'item';
+  icon?: IconName;
+}
+
+export type IconMapType = NodeType | 'group' | 'divider' | 'item';
```
casserni: Can we remove NodeType from IconMayType? I don't see it being used in any  styling so it shouldn't be needed

src/TableOfContents/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+  padding = '10',
+  title,
+  isOpen = false,
+  // tslint:disable-next-line: no-empty
+  onClose = () => {},
+  enableDrawer = true,
+}) => {
+  const contents = _contents;
+
+  // TODO (CL): Should we store expanded state in local storage?
```
lottamus:
```suggestion
```

src/TableOfContents/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+  const [expanded, setExpanded] = React.useState({});
+
+  const isMobile = useIsMobile(enableDrawer);
+
+  const comp = (
+    <div className={cn('TableOfContents bg-gray-1 dark:bg-transparent flex justify-end h-full border-r', className)}>
```
lottamus: Looks like we're not adding a border-r in Elements. Do you remember why we added one here? If the consumer needs a border, they can just pass it in via the className prop?

https://github.com/stoplightio/elements/blob/master/src/components/TableOfContents/index.tsx#L79

```suggestion
    <div className={cn('TableOfContents bg-gray-1 dark:bg-transparent flex justify-end h-full', className)}>
```

src/TableOfContents/index.tsx
```diff
@@ -0,0 +1,160 @@
+import { Button, Drawer, Icon } from '@blueprintjs/core';
+import cn from 'classnames';
+import * as React from 'react';
+import { useIsMobile } from '../_hooks/useIsMobile';
+import { ScrollContainer } from '../ScrollContainer';
+import { IContentsNode } from './types';
+
+export interface ITableOfContents {
+  // Precomputed list of nodes instead of items. This is useful if you want to customize the ordering of the tree.
```
lottamus:
```suggestion
```

PR #135, opened 2019-11-01
fix: code editor and viewer line numbers

lottamus:
- removes left padding when line numbers isn't enabled
- adds line numbers even when the textarea is empty
- adds highlighting to inline code
- adds memoization to prism highlight

image

### DIFF
--- a/src/CodeEditor/index.tsx
+++ b/src/CodeEditor/index.tsx
@@ -31,17 +31,19 @@ const CodeEditor = React.forwardRef<ReactSimpleCodeEditor, ICodeEditorProps>((pr
   } = props;
 
   // Highlight code on change
-  const highlight = React.useCallback(
-    (code: string) => (language ? highlightCode(code, language, showLineNumbers) : code),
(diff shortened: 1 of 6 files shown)


src/CodeEditor/index.tsx
```diff
@@ -31,17 +31,19 @@ const CodeEditor = React.forwardRef<ReactSimpleCodeEditor, ICodeEditorProps>((pr
...
+  ]);
 
   return (
     <ReactSimpleCodeEditor
       {...rest}
       // @ts-ignore FIXME type error
       ref={ref}
-      className={cn(className, 'bp3-code-editor')}
+      className={cn(className, 'bp3-code-editor', {
+        showLineNumbers,
```
marbemac: Camel case css classnames eh? 🤔

lottamus: _lazy CSS names_

lottamus: Done!

PR #142, opened 2020-03-13
feat: add RowsInput

wmhilton:
@P0lip (or anybody) why don't my `useState` hooks work in storybook? Also: is there any real benefit to this component?

@johnrazmus I'm honestly not sure at this point why we're doing this... but hey here's a PR.

I tried to port my GitSavedCredentials dialog to use it and now my v2 looks like this:
image

So I have no idea what's going on.

Edit: To clarify, I'm confused, not being negative.

### DIFF
--- a/src/RowsInput/index.tsx
+++ b/src/RowsInput/index.tsx
@@ -1,0 +1,40 @@
+import * as React from 'react';
+import { Button, ButtonGroup } from '..';
+
+export interface IRowInputProps<T> {
+  key: string;
+  row: T;
+  onRemove: () => void;
+  onChange: (row: T) => void;
+}
+
+export interface IRowsInputProps<T> {
+  rows: T[];
+  text: string;
+  onAddRow: () => void;
+  onRemoveRow: (i: number) => void;
+  onChangeRow: (i: number, row: T) => void;
+  RowInput: React.FunctionComponent<IRowInputProps<T>>;
+}
+
+export function RowsInput<T>({ rows, text, onAddRow, onRemoveRow, onChangeRow, RowInput }: IRowsInputProps<T>) {
+  const [rowsShiftedIndex, shiftRows] = React.useReducer(i => i + 1, 0);
+  return (
+    <>
+      <ButtonGroup>
+        <Button icon="plus" text={text} onClick={onAddRow} />
+      </ButtonGroup>
+      {rows.map((row, i) => (
+        <RowInput
+          key={`${i}-${rowsShiftedIndex}`}
+          row={row}
+          onRemove={() => {
+            onRemoveRow(i);
+            shiftRows();
+          }}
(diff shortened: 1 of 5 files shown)


src/RowsInput/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+  RowInput: React.FunctionComponent<IRowInputProps<T>>;
+}
+
+export function RowsInput<T>({ rows, text, onAddRow, onRemoveRow, onChangeRow, RowInput }: IRowsInputProps<T>) {
+  const [rowsShiftedIndex, shiftRows] = React.useReducer(i => i + 1, 0);
+  return (
+    <>
+      <ButtonGroup>
+        <Button icon="plus" text={text} onClick={onAddRow} />
+      </ButtonGroup>
```
lottamus: Instead of this add button, I suggest adding an empty row at the end of the list that calls `onAddRow` onFocus. You can see how we're doing this in the RequestMaker component for the parameters list: https://github.com/stoplightio/elements/blob/master/src/components/RequestMaker/Request/Parameters.tsx#L142-L148

lottamus:
> I tried to port my GitSavedCredentials dialog to use it and now my v2 looks like this:

@wmhilton v2.0 is using the `feat/platform-v2` branch: https://github.com/stoplightio/ui-kit/tree/feat/platform-v2

https://github.com/stoplightio/platform-internal/blob/v2/yalc.lock#L5

lottamus:
> I'm honestly not sure at this point why we're doing this

@wmhilton the goal is to provide a reusable "autogrowing" input list component that we can use in the git creds dialot, invite members modal, and request maker. I'm sure there will be plenty of other uses for this in the future too such as the OpenAPI params editor in Studio 🤔

wmhilton:
> > I'm honestly not sure at this point why we're doing this
> 
> @wmhilton the goal is to provide a reusable "autogrowing" input list component that we can use in the git creds dialog, invite members dialog, and request maker. I'm sure there will be plenty of other uses for this in the future too such as the OpenAPI params editor in Studio 🤔

But the problems that can be solved inside the component are…  pretty narrow. Basically just computing React’s key thing for array entries.

Since state management is so different depending on whether you’re using React.useState or MobX or OverMind the actual “how do I update an array” has to be delegated to the user.

There may yet be a point to it, if the goal is a consistent UX, but without a well defined story of what this component is supposed to be, I’m worried everybody will reach for it and then be like, “actually… it’s too limiting for my use case”. And it’ll become a monster component full of random options or a component that nobody uses and they just copy-paste it as a starting point.

P0lip:
> @P0lip (or anybody) why don't my useState hooks work in storybook? 

Why would we use them?

johnrazmus:
Not sure if this is needed anymore - we went a different route on the use-case for it

wmhilton:
Whew.

PR #144, opened 2020-04-02
chore: conventional displayName for `withErrorBoundary`

marcelltoth:
Makes `withErrorBoundary` follow the conventions for the `displayName` property.

https://reactjs.org/docs/higher-order-components.html#convention-wrap-the-display-name-for-easy-debugging

This allows easier debugging as it plays nicer with DevTools:
image

Even if I'm new to the projects and have no idea about what these are, I can immediately see that this is a `PageComponent` component passed into some `withErrorBoudary` HOC.

The explicit override opportunity is kept intact, this is just a nicer fallback.

### DIFF
--- a/src/withErrorBoundary.tsx
+++ b/src/withErrorBoundary.tsx
@@ -26,7 +26,8 @@ export function withErrorBoundary<T extends IErrorBoundary>(
       error: null,
     };
 
-    public static displayName = displayName;
+    public static displayName =
+      displayName || `withErrorBoundary(${WrappedComponent.displayName || WrappedComponent.name || 'Anonymous'})`;
 
     public componentDidUpdate(prevProps: Readonly<T>) {
       if (this.state.error !== null) {
--- a/src/styles/tailwind/_base.scss
+++ b/src/styles/tailwind/_base.scss
@@ -4924,10 +4924,12 @@
 }
 
 .flex {
+  display: -webkit-box;
   display: flex;
 }
 
 .inline-flex {
+  display: -webkit-inline-box;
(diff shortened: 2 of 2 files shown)


src/withErrorBoundary.tsx
```diff
@@ -26,7 +26,8 @@ export function withErrorBoundary<T extends IErrorBoundary>(
       error: null,
     };
 
-    public static displayName = displayName;
+    public static displayName =
```
P0lip, line 29: It's not the error boundary you are looking for.
Most, if not all, projects use [@stoplight/react-error-boundary](https://github.com/stoplightio/react-error-boundary) which already [supports](https://github.com/stoplightio/react-error-boundary/blob/master/src/withErrorBoundary.ts) the displayName correctly.

P0lip: If you have a component using the boundary exposed by ui-kit, simply convert it to use the new one.

marcelltoth: I'm coming here from `elements`, where this is being used still somewhere.

But if there's a separate repo for it, then why don't we:
1. Remove the duplicate implementation, make the `ui-kit` one an alias of the `react-error-boundary` one and deprecate it?
2. Remove the `ui-kit` boundary altogether in the next major.

P0lip: Yeah, setting an alias is OK, but you can also simply remove it.
I am aware this kind of a breaking change, but IIRC we don't use it anywhere else.

marcelltoth:
Closing in favor of #156

PR #145, opened 2020-04-07
feat: async validation support for FormInput

marcelltoth:
I added async yup schema support for `FormInput`.

This is something that can simplify our lives many times when implementing async validations for hubs.

I additionally added a story to test and demonstrate the behavior, and improved typings a little.

### DIFF
--- a/src/__stories__/FormInput/index.tsx
+++ b/src/__stories__/FormInput/index.tsx
@@ -1,0 +1,45 @@
+import * as React from 'react';
+import * as yup from 'yup';
+
+import { withKnobs } from '@storybook/addon-knobs';
+import { storiesOf } from '@storybook/react';
+
+import { FormInput } from '../../';
+
+storiesOf('FormInput', module)
+  .addDecorator(withKnobs)
+  .add('with synchronous validation', () => (
+    <div className="p-40">
+      <FormInputContainer schema={exampleSyncSchema} />
+    </div>
+  ))
+  .add('with async validation', () => (
+    <div className="p-40">
+      <FormInputContainer schema={exampleAsyncSchema} />
+    </div>
+  ));
+
+const FormInputContainer: React.FC<{ schema: yup.StringSchema }> = ({ schema }) => {
+  const [value, setValue] = React.useState('');
+
+  return (
+    <FormInput
+      value={value}
+      onChange={(e: React.ChangeEvent<HTMLInputElement>) => setValue(e.currentTarget.value)}
+      schema={schema}
+    />
+  );
+};
+
+const exampleSyncSchema = yup
+  .string()
+  .min(3, 'waaay too short')
+  .max(5, 'waaay too long');
+
+const exampleAsyncSchema = yup.string().test('oddchars', 'The number of characters must be odd', value => {
+  return new Promise(resolve => {
+    setTimeout(() => {
+      resolve(value.length % 2 === 1);
+    }, 1000);
+  });
+});
--- a/src/_hooks/useValidateSchema.ts
+++ b/src/_hooks/useValidateSchema.ts
@@ -1,3 +1,4 @@
+import { isEqual } from 'lodash';
 import * as React from 'react';
 import * as yup from 'yup';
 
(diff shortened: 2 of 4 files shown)


XVincentX:
Can do quickly do an example for Hubs? For example, hub uniqueness verification?

marcelltoth:
> 
> Can do quickly do an example for Hubs? For example, hub uniqueness verification?

```typescript
const slugSchema = React.useMemo(() => 
  yup.string()
    .min(3)
    .test('unique', 'this slug is used by another hub already', value => {
      const { data } = await urqlClient.query(HubBySlugDocument, { slug, withConfigs: false }).toPromise();
      return data.hubs.length;
    })
, [urqlClient]);

return <FormInput ... schema={slugSchema} />
```

src/_hooks/useValidateSchema.ts
```diff
@@ -1,16 +1,43 @@
...
+      })
+      .catch(e => {
+        setErrors(e.errors || unknownError);
+      });
+  }, [memoizedSchema, value, abortEarly, recursive, memoizedContext]);
+
+  return errors;
+}
+
+function useDeepCompareMemoize<T>(value: T) {
```
XVincentX, line 35: Not too sure I understand why we need this; doesn't React.memo do the same thing?

marcelltoth: `React.memo` is shallow. In an ideal world, `schema` is memoized, but there are probably people who will call the component as:
```typescript
// in a component
const schema = yup.string().something();
return <FormInput schema={schema} />;
```
which is a new schema each time. We don't want to run the effect again in these cases, so a deep compare takes care of that.

XVincentX: Hmm, I am not too sure the component should handle bad usages, I would prefer to document the behaviour and discipline the user. I do not know if this is a pattern or not but it's not blocking this though.

src/__stories__/FormInput/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+const exampleSyncSchema = yup
+  .string()
+  .min(3, 'waaay too short')
+  .max(5, 'waaay too long');
+
+const exampleAsyncSchema = yup.string().test('oddchars', 'The number of characters must be odd', value => {
+  return new Promise(resolve => {
+    setTimeout(() => {
+      resolve(value.length % 2 === 1);
```
XVincentX, line 42: What happens if the promise gets rejected?

marcelltoth: By the docs you are not supposed to provide anything that can reject:
>  To make a test async return a promise that resolves true or false or a ValidationError.
https://github.com/jquense/yup#mixedtestname-string-message-string--function-test-function-schema

But I haven't tried, should I?

src/_hooks/useValidateSchema.ts
marcelltoth: In general I agree with you completely, but in this case I was very afraid that the above pattern is being used, and I cause a breaking change with this PR.

But I can take it out and see, should I?

src/__stories__/FormInput/index.tsx
XVincentX: Hmm,  if that's the case, I am wondering wether we shouldn't catch the promise error here and resolve it with a `false`

marcelltoth: This is a story and an example timeout promise. What do you mean by `here`? In the hook?

XVincentX: Yes, in the hook. What do you think?

src/_hooks/useValidateSchema.ts
XVincentX: > In general I agree with you completely, but in this case I was very afraid that the above pattern is being used

Then we fix it in the codebase I'd say 😄

src/__stories__/FormInput/index.tsx
marcelltoth: I checked the behavior now.
`yup` really does not catch the error, so it falls through to our catch handler:
```typescript
.catch(e => {
  setErrors(e.errors || unknownError);
});
```

This usually ends up as "Invalid input" displayed in the UI.

In the hook I don't have access to the actual validation promise (in a sense that I could catch the error and turn it into a regular `false`.)

Nether do I think it would be good practice on our part to override how `yup` works. The user is creating `yup` schemas, they should create well behaving ones.

What we *can* do is to make sure that our hook handles these unknown errors, and I just did that: it does. One thing to consider from an UX perspective is to rename the "Invalid input" fallback to something like "Unknown error", that conveys that it's possibly a programming error, not something on the user's side.

src/_hooks/useValidateSchema.ts
marcelltoth: I'm cool with that. I removed it!

XVincentX:
@marcelltoth Can you remove the CodeClimate thing in CircleCI? We are not using it anymore

src/__stories__/FormInput/index.tsx
marcelltoth: Done

PR #146, opened 2020-04-16
feat: debounced validation for `FormInput` / `FormButton`

marcelltoth:
This is based on @lottamus's PR feedback for https://github.com/stoplightio/platform-internal/pull/1748

## Summary

Debounces the yup schema validation `FormInput` and `FormButton` have.

## UI changes

These are opinionated, let me know if you want to change anything.

- A valid input still does not show any icon.
- The invalid icon was changed to include an exclamation mark in the middle to be more consistent with the _stale/validating_ icon (see below). 
image
- If there _was_ an error during the latest validation, but the input changed since (_stale/validating_), this icon is shown:
image
- If the last state was _correct_ but the input changed since, we don't show anything. A flashing icon would look bad IMO, so unless we add an icon for the valid state, this is probably nicer behavior.

I also fixed the vertical alignment of the icons so they are always in the middle.

### `useValidateSchema`

`useValidateSchema` now returns an object instead of a single string array. This is not a big deal as the hook is considered internal, I guess.

### `FormButton`

There are some "breaking" changes here. They improve type safety, neither affects `platform-internal`, and this is `beta`, but still, let me know if you don't like it here.

- `IFormButtonProps`, aka `IFormButton` is not exported anymore. There's no point. 
  
  ```typescript
  type IFormButtonProps = React.ComponentProps<typeof FormButton>;
  ```
  and that's it, if anyone needs it. But we never used it in the first place.
- `FormButton` is generic. This is to remove the `any`s from the props, and to make sure that `typeof data` aligns with the schema. I can't do `<FormButton data="abc" schema={yup.number()}/>` anymore.
- `FormButton`'s `onClick` lost it's attribute. It would not work with the new typing, and also there's absolutely 0 point, as the parameter passed was always simply the `data` prop back.

### DIFF
--- a/src/FormButton/index.tsx
+++ b/src/FormButton/index.tsx
@@ -19,11 +19,11 @@ type OwnProps<T> = {
 type IFormButtonProps<T> = Omit<IButtonProps, 'type' | 'onClick'> & OwnProps<T>;
 
 function FormButton<T>({ schema, data, loading, disabled, onClick, ...buttonProps }: IFormButtonProps<T>) {
-  const { errors, isStale } = useValidateSchema(schema, data);
+  const { errors, isValidating } = useValidateSchema(schema, data);
 
   return (
     <Button
-      disabled={errors.length > 0 || isStale || loading || disabled}
+      disabled={errors.length > 0 || isValidating || loading || disabled}
       loading={loading}
       onClick={onClick}
       {...buttonProps}
--- a/src/FormButton/index.tsx
+++ b/src/FormButton/index.tsx
@@ -25,12 +25,12 @@ const FormButton: React.FunctionComponent<IFormButton> = ({
   onClick = noop,
   ...buttonProps
 }) => {
-  const { errors } = useValidateSchema(schema, data);
+  const { errors, isStale } = useValidateSchema(schema, data);
   const handleClick = React.useCallback(() => onClick(data), [onClick, data]);
 
   return (
     <Button
-      disabled={errors.length > 0 || loading || disabled}
+      disabled={errors.length > 0 || isStale || loading || disabled}
       loading={loading}
       onClick={handleClick}
       {...buttonProps}
--- a/src/FormButton/index.tsx
+++ b/src/FormButton/index.tsx
@@ -25,7 +25,7 @@ const FormButton: React.FunctionComponent<IFormButton> = ({
   onClick = noop,
   ...buttonProps
 }) => {
-  const errors = useValidateSchema(schema, data);
+  const { errors } = useValidateSchema(schema, data);
   const handleClick = React.useCallback(() => onClick(data), [onClick, data]);
 
   return (
--- a/src/FormButton/index.tsx
+++ b/src/FormButton/index.tsx
@@ -19,7 +19,7 @@ type OwnProps<T> = {
 type IFormButtonProps<T> = Omit<IButtonProps, 'type' | 'onClick'> & OwnProps<T>;
 
 function FormButton<T>({ schema, data, loading, disabled, onClick, ...buttonProps }: IFormButtonProps<T>) {
-  const { errors, isValidating } = useValidateSchema(schema, data);
+  const [{ errors, isValidating }] = useValidateSchema(schema, data);
 
   return (
     <Button
--- a/src/FormButton/index.tsx
+++ b/src/FormButton/index.tsx
@@ -1,4 +1,3 @@
-import { noop } from 'lodash';
 import * as React from 'react';
 // TODO: should probably use ajv and json schema
 import * as yup from 'yup';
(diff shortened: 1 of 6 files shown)


src/_hooks/useValidateSchema.ts
```diff
@@ -8,7 +9,10 @@ export function useValidateSchema<T>(
   schema?: yup.Schema<T>,
   value?: T,
   { abortEarly, recursive }: yup.ValidateOptions = {},
-): string[] {
+) {
+  const [debouncedValue] = useDebounce(value, 500);
```
lottamus, line 13: Maybe we can pass the debounce timer in as an option to FormInput?

```suggestion
  const [debouncedValue] = useDebounce(value, debounce);
```

src/FormInput/index.tsx
```diff
@@ -89,7 +89,19 @@ const FormInputValidation: React.FunctionComponent<IFormInputValidationProps> =
...
 }) => {
-  const errs = useValidateSchema(schema, value, { abortEarly: false }).concat(errors);
+  const { errors: validationErrors, isStale } = useValidateSchema(schema, value, { abortEarly: false });
+  const errs = [...validationErrors, ...errors];
+
+  if (isStale && errs.length) {
+    return (
+      <Tooltip content="Validating..." position={Position.BOTTOM_RIGHT}>
+        <div tabIndex={-1} style={{ height: iconHeight[size] }} className="mr-2">
+          <Icon icon="outdated" iconSize={size === 'small' ? 12 : Icon.SIZE_STANDARD} intent="danger" />
```
lottamus, line 99: What do you think about using a red [spinner](https://blueprintjs.com/docs/#core/components/spinner) here while we're validating instead of the outdated icon? I'm afraid the outdated icon might be confusing

src/_hooks/useValidateSchema.ts
```diff
@@ -17,14 +21,14 @@ export function useValidateSchema<T>(
...
         setErrors(noError);
       })
       .catch(e => {
         setErrors(e.errors || unknownError);
       });
-  }, [schema, value, abortEarly, recursive]);
+  }, [schema, debouncedValue, abortEarly, recursive]);
 
-  return errors;
+  return { errors, isStale };
```
XVincentX, line 33: Should this maybe be an array, in a similar way other hooks do? `[errors, isStale]`?

XVincentX: @lottamus ^

lottamus: Is there a benefit? What would that look like?

```suggestion
  return [{ errors, isStale }];
```

src/FormInput/index.tsx
XVincentX: 1. I agree that the spinner represents better the operation in progress.
2. For async validation I think the spinner should always appear, even if the current input is valid

src/_hooks/useValidateSchema.ts
XVincentX: It would be similar to state `[state,setState] = React.useState('')` — it's just a style thing though

src/FormButton/index.tsx
```diff
@@ -11,34 +10,28 @@ import { Button, IButtonProps } from '../index';
...
+function FormButton<T>({ schema, data, loading, disabled, onClick, ...buttonProps }: IFormButtonProps<T>) {
+  const [{ errors, isValidating }] = useValidateSchema(schema, data);
 
   return (
     <Button
-      disabled={errors.length > 0 || loading || disabled}
+      disabled={errors.length > 0 || isValidating || loading || disabled}
       loading={loading}
-      onClick={handleClick}
+      onClick={onClick}
```
lottamus, line 28: We could leave this out now and use the typings from `IFormButtonProps`. What do you think?

```suggestion
```

src/FormButton/index.tsx
```diff
@@ -11,34 +10,28 @@ import { Button, IButtonProps } from '../index';
...
  */
-interface IFormButton extends Omit<IButtonProps, 'type' | 'onClick'> {
-  schema?: yup.Schema<any>;
-  data?: any;
-  onClick?: (data: any) => void;
-}
+type OwnProps<T> = {
+  schema?: yup.Schema<T>;
+  data?: T;
+  onClick?: () => void;
```
lottamus, line 16: If we use the IFormButtonProps onClick typing we don't need this

```suggestion
```

src/FormButton/index.tsx
XVincentX: Ok, but I still need to bind the event handler, so this line has to stay, right?

lottamus: Nope, we can pass it straight through with the other button props being destructured on the line below. Just need to remove it from the react props above and remove the onClick typing: https://github.com/stoplightio/ui-kit/pull/146/files/794831b3eb395472ed38e6129959670f333ebead#diff-d2297309074b8bcc21abf16d7bd22309R21

Main use case is to allow handling the click event

```tsx
<FormButton
  onClick={e => e.preventDefault()}
/>
```

XVincentX: Ah I understand what you mean now, ok.

PR #147, opened 2020-04-16
fix: object validation for `useValidateSchema` & story for FormButton

marcelltoth:
Currently `FormButton` does not work with object `data` and `schema`, which makes it quite useless.

The problem is that for some reason if we pass a configuration object to `yup`, even `undefined` properties of said object override defaults. (They act as false.) 

This means we were effectively always passing `abortEarly: false, recursive: false`. I fixed the default for `abortEarly`, and I removed the option for `recursive` as neither did we ever use it, nor can I think of a situation where it can be useful. This is an internal hook, so the change is not breaking.

I also added a story to highlight the problem, that takes up the majority of lines added in this PR.

### DIFF
--- a/src/_hooks/useValidateSchema.ts
+++ b/src/_hooks/useValidateSchema.ts
@@ -17,7 +17,7 @@ export function useValidateSchema<T>(
       return;
     }
     schema
-      .validate(value, { strict: true, abortEarly }) // to avoid taking a useEffect dependency on validateOpts that is an object
+      .validate(value, { strict: true, abortEarly: abortEarly ?? true }) // to avoid taking a useEffect dependency on validateOpts that is an object
       .then(() => {
         setErrors(noError);
       })
--- a/src/_hooks/useValidateSchema.ts
+++ b/src/_hooks/useValidateSchema.ts
@@ -7,7 +7,7 @@ const unknownError = ['Unknown error'];
 export function useValidateSchema<T>(
   schema?: yup.Schema<T>,
   value?: T,
-  { abortEarly, recursive }: yup.ValidateOptions = {},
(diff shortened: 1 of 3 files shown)


src/_hooks/useValidateSchema.ts
```diff
@@ -17,14 +17,14 @@ export function useValidateSchema<T>(
       return;
     }
     schema
-      .validate(value, { strict: true, abortEarly, recursive }) // to avoid taking a useEffect dependency on validateOpts that is an object
+      .validate(value, { strict: true, abortEarly: abortEarly ?? true }) // to avoid taking a useEffect dependency on validateOpts that is an object
```
P0lip, line 20: This should do the same assuming abortEarly is a boolean.
```suggestion
      .validate(value, { strict: true, abortEarly: abortEarly !== false }) // to avoid taking a useEffect dependency on validateOpts that is an object
```

lottamus:
> do we actually use recursive anywhere or not?

Not that I'm aware of but is there any reason to remove it as an option?

XVincentX:
> The problem is that for some reason if we pass a configuration object to yup, even undefined properties of said object override defaults. (They act as false.)

This is the behaviour of `Object.assign` (which I am assuming yup might be using). Lodash's `merge` method does not override if the proposed argument is undefined.

src/_hooks/useValidateSchema.ts
marcelltoth: It may just be my C# background (where we've had `?.` and `??` for ages), but the nullish coalescing version is clearer for me. I sometimes call it the "fallback operator". I can understand yours as well of course, but it requires some thinking on how the thing works and what the default is.

PR #149, opened 2020-04-17
Fix: use ES modules in the output

marcelltoth:
Essentially the replica of https://github.com/stoplightio/elements/pull/339 for `ui-kit`.

### DIFF
--- a/tsconfig.build.json
+++ b/tsconfig.build.json
@@ -9,7 +9,6 @@
   "exclude": ["**/__*__/**"],
 
   "compilerOptions": {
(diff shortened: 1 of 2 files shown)


marcelltoth:
Yalced into the platform along with the new `elements`:
image

marcelltoth:
> 
> 
> This is not a fix, it's likely a breaking change @marcelltoth

True, but it's targeted at _beta_ for that reason.

XVincentX:
Ah true, I didn't notice that. Kaboom!

PR #159, opened 2020-05-15
feat: properly type RowRenderer

marcelltoth:
hint: It's still a mess

This is relevant because of https://github.com/stoplightio/platform-internal/issues/2527

That error is a result of us using an any-typed `rowRenderer`:
```ts
const rowRenderer = React.useCallback<any>(
```
https://github.com/stoplightio/elements/blob/beta/src/components/TableOfContents/index.tsx#L19

It turns out it is incredibly hard to correctly get the type of the callback, and having that `getProps: typeof computeTableOfContentsItemProps` is also prone to accidental breaking changes.

I pulled out the type of the rowRenderer to an explicitly defined type, and inlined the `typeof computeTableOfContentsItemProps` definition to make it a stable API.

IMO this renderer signature is still a mess, but I intended not to change it in this PR, just type it more explicitly.

### DIFF
--- a/src/TableOfContents/index.tsx
+++ b/src/TableOfContents/index.tsx
@@ -35,16 +35,24 @@ export type ITableOfContentsLink = TableOfContentsItem & {
   isExternalLink?: boolean;
 };
 
+export type RowRendererType<T extends TableOfContentsItem> = (props: {
+  item: T;
+  key: number | string;
+  getProps: (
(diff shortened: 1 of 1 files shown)


marcelltoth:
> 
> 
> Oops. I thought I approved this yesterday but I must have lost the tab.

No rush on this one. 🏖️ 🕶️

PR #161, opened 2020-05-19
feat(table-of-contents): add data-test attribute

marcelltoth:
Related to stoplightio/platform-internal#2527

While the PR for that issue (stoplightio/platform-internal#2537) was in review, a new prop was added to the TableOfContents component in `ninja`. stoplightio/platform-internal#2332

In this PR I migrate that prop too.

Two changes I made:
- I named the prop `data-test` instead of `dataTest` so it is consistent with the native `data-test` prop on HTML elements.
- I added a short documentation sentence.

When I pull this into stoplightio/platform-internal#2537 I'll update the e2e test to reflect the prop name change.

### DIFF
--- a/src/TableOfContents/index.tsx
+++ b/src/TableOfContents/index.tsx
@@ -50,6 +50,11 @@ export interface ITableOfContents<T extends TableOfContentsItem = TableOfContent
   padding?: string;
   className?: string;
 
+  /**
+   * HTML data-test attribute to be set on the container div.
+   */
+  'data-test': string;
+
   // force items to render with active or selected if either is true.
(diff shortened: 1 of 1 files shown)


src/TableOfContents/index.tsx
```diff
@@ -50,6 +50,11 @@ export interface ITableOfContents<T extends TableOfContentsItem = TableOfContent
   padding?: string;
   className?: string;
 
+  /**
+   * HTML data-test attribute to be set on the container div.
+   */
+  'data-test'?: string;
```
mallachari, line 56: I was about to say that it should be optional but you were quicker ;)

marcelltoth:
> 
> 
> 👍

Thank you!

PR #163, opened 2020-05-21
feat(Table of Contents): auto expand to show active items

marcelltoth:
The `TableOfContents` component will auto expand and scroll to make sure it shows items that are `isActive: true`, by default.

Will be used to resolve stoplightio/platform-internal#2308

The expand part works, scroll functionality to be done.

### DIFF
--- a/src/TableOfContents/index.tsx
+++ b/src/TableOfContents/index.tsx
@@ -1,5 +1,6 @@
 import { Button, Drawer, InputGroup } from '@blueprintjs/core';
 import cn from 'classnames';
+import { flatMap } from 'lodash';
 import * as React from 'react';
 
 import { FAIcon, FAIconProp } from '../FAIcon';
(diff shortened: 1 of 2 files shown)


marcelltoth:
Okay, so the scrolling mechanism turns out to be next to impossible without making a breaking change because of the unconventional way the `rowRenderer` / `DefaultRow` is designed.

The problem is not actually with scrolling the node into view, I have achieved that quickly. The problem is to make sure we are only doing that on the initial render, not on every update.

The problem with that is this `DefaultRow` is weird. It is by signature a component, but it is generated on the fly, for **each render** of **each row separately**.

This means that this component (and recursively its children including `TableOfContentsRowInner`) get **remounted each time**, resulting in
1. bad performance
2. the component not being able to tell if it's a mount or an update. (Everything is a mount.)

I suggest refactoring the solution to a much simpler design, preferably befor ewe take this out of beta:
- `DefaultRow` would be a static exported component that describes how a ToC item looks like by default.
- the default `rowRenderer` would just do `props => <DefaultRow {...props}/>`.
- If a consumer (like `ninja`) would want to alter the `rowRenderer` while still incorporating the original implementation they could do (example): `rowRenderer={props => <a href={props.to}><DefaultRow {...props }/></a>}`
- This in turn makes `rowRenderer` act like a component so we could consider taking a *Component* instead of a *renderer*, just like how `react-select`'s pretty amazing extensibility model works: https://react-select.com/components

src/TableOfContents/index.tsx
```diff
@@ -382,12 +395,24 @@ const TableOfContentsItemInner = ({
 /**
  * Traverses contents backwards to find the first index with a lower depth
  */
-function findParentIndex(currentDepth: number, contents: TableOfContentsItem[]) {
+function findParentIndex(currentDepth: number, contents: TableOfContentsItem[]): number {
   for (let index = contents.length - 1; index >= 0; index--) {
     if (contents[index].depth === currentDepth - 1) {
-      return String(index);
+      return index;
```
mallachari, line 401: Why was it string before? Kind of weird.

marcelltoth: Yes, dunno.

marcelltoth:
@mallachari @stoplightio/void-crew What do you think about my proposal? Should I create a chore issue?

mallachari:
@marcelltoth I'm all for your proposal. If we can have it act like normal component that updates instead of being remounted then that's a good way to go.

PR #166, opened 2020-05-27
fix: XSS vulnerability in CodeViewer

### LINKED ISSUE #164: [XSS] CodeViewer interprets HTML
**Describe the bug**

A CodeViewer component will interpret the contents instead of simply rendering it.

**To Reproduce**

1. Instantiate a CodeViewer with the following props:
```json
{
  "className": "overflow-auto MV_block",
  "value": "<div>Hello world!</div>\n<img src=\"asd\" onerror=\"alert('dont do this')\" />",
  "language": null,
  "showLineNumbers": false
}
```
2. Render it.
3. Cry.

**Expected behavior**
It would render and highlight the HTML.

**Screenshots**
This is from the markdown-viewer storybook:
image

wmhilton:
fix #164 

Stealing heavily from https://github.com/stoplightio/markdown-viewer/pull/50

Question: Should we export `createSanitize` so that `markdown-viewer` can import it from `ui-kit`?

image


src/CodeViewer/index.tsx
```diff
@@ -29,7 +42,11 @@ const CodeViewer: React.FunctionComponent<ICodeViewerProps> = ({
 }) => {
   const lang = (language && languageMaps[language]) || language || '';
 
-  const code = React.useMemo(() => highlightCode(value, lang, showLineNumbers), [value, lang, showLineNumbers]);
+  const code = React.useMemo(() => sanitizer.sanitize(highlightCode(value, lang, showLineNumbers), sanitizeConfig), [
+    value,
+    lang,
+    showLineNumbers,
+  ]);
```
marcelltoth, line 49: As far as I understand `highlightCode` returns a HTML string that contains the formatting. Just make sure we don't strip the highlighting. This is a more complicated issue then the one in markdown-viewer.

wmhilton: Yup. But the highlighted code only inserts `<span>`s. So I just set `ALLOWED_TAGS: ['span']` and that seems to work:

image

marcelltoth:
I broke it! 😎 See extended story.

image

wmhilton:
Ugh. This is proving... surprisingly tricky.

lottamus:
Alternatively we could look at using something like https://github.com/conorhastings/react-syntax-highlighter

wmhilton:
Hmmmm that might be a good fit. They even have an option to use Prism languages: https://github.com/conorhastings/react-syntax-highlighter/blob/master/README.md#light-build

I'll try it out, see if it avoids the XSS issues. It says it doesn't use `dangerouslySetHTML` so that could work.

marcelltoth:
> 
> 
> Alternatively we could look at using something like https://github.com/conorhastings/react-syntax-highlighter

Be careful, `react-syntax-highlighter`'s Prism is some 138KB mingzip, vs prismjs that is 5.6

P0lip:
> e careful, react-syntax-highlighter's Prism is some 138KB mingzip, vs prismjs that is 5.6. 

I suppose that the size you mentioned does not include all languages Prism has to offer.
Prismjs does not register all languages by default, hence we need to load them manually.
Either way, even with the relevant languages included, the size is likely to be closer to like 15KB than 138KB :laughing:

P0lip:
FWIW, we can simply checkout `33b3809fda3f2cd6d8a4b91f0e1b12f67e320645` :see_no_evil:.

marcelltoth:
> 
> 
> > e careful, react-syntax-highlighter's Prism is some 138KB mingzip, vs prismjs that is 5.6.
> 
> I suppose that the size you mentioned does not include all languages Prism has to offer.
> Prismjs does not register all languages by default, hence we need to load them manually.
> Either way, even with the relevant languages included, the size is likely to be closer to like 15KB than 138KB 😆

Yes of course. My point is, if we include the non-Light version of that lib, that brings in all the languages.

P0lip:
Closing in favor of #168

PR #168, opened 2020-06-02
fix: restore older impl of CodeViewer

### LINKED ISSUE #164: [XSS] CodeViewer interprets HTML

P0lip:
Brings back older implementation of CodeViewer that didn't use `dangerouslySetInnerHTML` and therefore was less prone to XSS.

Fixes #164


marcelltoth:
Honestly I'm not a great fan of doing this by hand. It's a 1000+ line PR (I'm aware most of that is tests and fixtures, but still) and as far as I can tell, it's building exactly the same thing as `react-syntax-highlighter/prism`, a graphical frontend for `refractor`.

Does this solution offer something extra that we couldn't get by using the library? If not I'd prefer not to maintain 100s of lines of code & tests, we should just use the lib.

P0lip:
> I love the AST approach, but honestly I'm not a great fan of doing this by hand. It's a 1000+ line PR (I'm aware most of that is tests and fixtures, but still) and as far as I can tell, it's building exactly the same thing as react-syntax-highlighter/prism or react-refractor, a graphical frontend for refractor.

> Does this solution offer something extra that we couldn't get by using the library? If not I'd prefer not to maintain 100s of lines of code & tests, we should just use the lib.

It was just simpler to restore the older code that worked well than to adjust some existing library.
Moreover, this approach has one huge benefit - we can execute the  "processing code" in a worker.
We're going to need it eventually to fix issues like https://github.com/stoplightio/platform-internal/issues/2643.
My initial plan was to do it as a part of this PR, but I realized it'd cause unnecessary scope creep and in general I thought it was something we might potentially not want at this point.

marcelltoth:
> we can execute the "processing code" in a worker.

You convinced me here. 👍

PR #171, opened 2020-06-05
feat: toc hidden item action

mallachari:
Needed for: https://github.com/stoplightio/platform-internal/issues/2823 as we started to use this element for reporting.

Adds optional hidden property to item action button. We could reuse `isActive` but that may have also different purpose.

### DIFF
--- a/src/TableOfContents/index.tsx
+++ b/src/TableOfContents/index.tsx
@@ -26,6 +26,7 @@ export type TableOfContentsItem = {
     icon?: FAIconProp;
     name?: string;
     isActive?: boolean;
+    hidden?: boolean;
     onClick: any;
   };
 };
@@ -359,7 +360,9 @@ const TableOfContentsItemInner = ({
         onClick={showSkeleton ? undefined : item.action.onClick}
         active={item.action.isActive}
         intent={item.action.isActive ? 'primary' : undefined}
-        className="ml-2"
+        className={cn('ml-2', {
+          invisible: item.action.hidden,
+        })}
         minimal
         small
       />


mallachari:
> Is there a meaningful difference between `{..., action: {..., hidden: true}}` and `action: undefined`?

Yes as we are currently using that action to send reports for node opening. It was added few days ago: https://github.com/stoplightio/platform-internal/blob/v2/packages/ninja/src/screens/Workspace/Sidebar/Project.tsx#L93
So we want to keep the handler but don't show the button.

marcelltoth:
> Yes as we are currently using that action to send reports for node opening. It was added few days ago: https://github.com/stoplightio/platform-internal/blob/v2/packages/ninja/src/screens/Workspace/Sidebar/Project.tsx#L93
> So we want to keep the handler but don't show the button.

But isn't the button the definition of `action`? Why don't you just overwrite the `onClick`? Or if that doesn't work, add a callback prop to the ToC to signal when a node was expanded.

mallachari:
You may be right. I went with action as it was already done this way. I'll see the alternatives. Thanks!

marcelltoth:
> 
> 
> You may be right. I went with action as it was already done this way. I'll see the alternatives. Thanks!

👍 Looking at it further I don't believe the current code even does what they intended. It only tracks clicks on that dark gray box your issue is about. It will not track if I click anywhere else to expand the row.

So you are actually fixing two problems at once, #2823, and the bug in the reporting-system.

mallachari:
>  Looking at it further I don't believe the current code even does what they intended. It only tracks clicks on that dark gray box your issue is about. It will not track if I click anywhere else to expand the row.
> 
> So you are actually fixing two problems at once, #2823, and the bug in the reporting-system.

Actually surprisingly it did work when clicked outside of it - I tried that as I thought the same.

Anyway, it can work with onClick on the item in the end so I'm closing this one. Thanks again!

marcelltoth:
Wow, that's surprising. It really shouldn't. 😮 Funnily that might signal a bug around the action's `onClick` handling. (One that worked in our favor this time.)

PR #172, opened 2020-06-09
fix(code-viewer): restore previous css

P0lip:
image

### DIFF
--- a/src/styles/components/Code/_base.scss
+++ b/src/styles/components/Code/_base.scss
@@ -36,17 +36,19 @@
   }
 
   .line-number {
+    padding-left: 45px;
     position: relative;
-    width: 35px;
-    display: inline-block;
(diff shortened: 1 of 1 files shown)


marbemac:
Can't tell from the screenshot, but does it have some sort of opacity set? It looks somewhat transparent.

src/styles/components/Code/_base.scss
```diff
@@ -36,17 +36,19 @@
...
+    display: block;
 
     &::before {
-      content: '';
       counter-increment: line;
       content: counter(line);
+      display: inline-block;
+      position: absolute;
+      left: 5px;
+      opacity: 0.4;
```
P0lip, line 50: @marbemac  Yeah, the opacity is set here.
Shall we adjust it or get rid of it?

PR #173, opened 2020-06-17
feat(toc): scroll the active item into view on mount

marcelltoth:
Fixes the scroll part of stoplightio/platform-internal#2308

### DIFF
--- a/src/__stories__/TableOfContents/index.tsx
+++ b/src/__stories__/TableOfContents/index.tsx
@@ -17,11 +17,7 @@ const styles = {
 storiesOf('TableOfContents', module)
   .addDecorator(withKnobs)
   .add('studio /w custom RowComponent', () => {
-    return (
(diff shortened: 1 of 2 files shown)


src/__stories__/TableOfContents/index.tsx
```diff
@@ -52,8 +48,42 @@ const MobileStory = () => {
...
+
+const RowComponent: RowComponentType<ITableOfContentsLink> = props => {
+  const { setPath } = React.useContext(NavigationContext);
+
+  const handleClick = React.useCallback(
+    (e: React.MouseEvent) => {
+      if (props.item.to) {
+        setPath(props.item.to);
+      }
+      e.preventDefault();
```
mallachari, line 79: Yey! So much better with it!

marcelltoth:
Thanks for the review @mallachari !

Issue #165, opened 2020-05-27
Support Tree shaking 


marcelltoth:
**User story.**
As a developer, I can use the pieces of `ui-kit` without bloating my package by including the entire bundle.

**Is your feature request related to a problem?**
See stoplightio/elements#376 for example. Elements uses parts of `ui-kit`, then the elements bundle gets very large as it imports all of `@blueprintjs/core`, even parts that it doesn't need.

Ninja has this problem, to a lesser degree. Lesser because I assume we use almost the entire `ui-kit` in ninja, but probably still not 100%.

**Describe the solution you'd like**
Implement dual ESM + CJS packaging. yalc into ninja to see if SSR support (CJS) is still intact.

**Additional context**
I have created an example project to demonstrate the issue: https://github.com/marcelltoth/bundle-size-test
This is a simple CRA template that includes a single button from `ui-kit`.

Using the current *beta* the resulting bundle looks like this:
![image](https://user-images.githubusercontent.com/543372/83013813-4c729300-a026-11ea-8785-ff85e6b0f46e.png)

I experimented with packing the same `ui-kit` as an ESM. Looks much better (35% improvement):
![image](https://user-images.githubusercontent.com/543372/83015475-2bf80800-a029-11ea-9add-92278163e826.png)

The second one still has the problem of including all blueprint icons, that's an open issue at https://github.com/palantir/blueprint/issues/2193

marcelltoth:
We should do a followup in elements once this is done.

P0lip:
Misclicked.

Actually, we may want to consider upgrading to Webpack 5 as well and see what the gains are.
It's still in beta, but according to them, cjs tree-shaking is supported to some extent.

PR #174, opened 2020-06-29
perf(code-viewer): pre-render only relevant parts

P0lip:
_Fixes https://github.com/stoplightio/platform-internal/issues/2643

code-viewer

(note, the above video was captured with windowing enabled)

besides tests that need to be updated, the remaining question is how we should bundling
I have a couple of ideas:
- inline the worker same as we do in jsv@2
- use worker-loader and rely on end-app to use appropriate bundler (webpack)
- follow the monaco-editor path

@lottamus @marbemac @stoplightio/void-crew 
whatcha think?


wmhilton:
> the remaining question is how we should bundling

I would never do the `monaco-editor` path again frankly. The only reason we did it that way is because Monaco is weird and has its own loader.

So I'd say it's between the worker-loader approach and the inline approach (which also uses worker-loader I assume? but during the build of `ui-kit` rather than during the build of `ninja`?).

As long as it doesn't kill loading time, I'd do the inline approach. It's the most portable and least fragile.

You forgot the 4th approach, the one that Studio uses, where we pass in a `createWorker` function prop. 😆 Yeah, I honestly am not a fan of that one either - I only did it that way because of Electron.

P0lip:
> So I'd say it's between the worker-loader approach and the inline approach (which also uses worker-loader I assume? but during the build of ui-kit rather than during the build of ninja?).

Inling proved to be very portable - this has been battled tested in jsv@2, with that difference that we used rollup there, but nothing stopping us from using webpack here.
I'll go down this path then.

src/CodeViewer/index.tsx
```diff
@@ -43,21 +81,51 @@ const CodeViewer: React.FunctionComponent<ICodeViewerProps> = ({
...
-        [`${Classes.CODE_EDITOR}--inline`]: inline,
-        [`${Classes.CODE_EDITOR}--line-numbers`]: showLineNumbers,
-      })}
-      {...rest}
-    >
-      {markup ? markup.map(astToReact()) : value}
-    </pre>
+    <AutoSizer>
+      {({ height, width }) => (
+        <FixedSizeList
```
P0lip, line 100: Note - the code will not be copy-able anymore, because we'll only render a small chunk of it.
Is this okay?

P0lip: Got rid of it eventually, since search would be broken either, and this is likely to be annoying.

marbemac:
> use worker-loader and rely on end-app to use appropriate bundler (webpack)

I think there are interesting learnings in general to be had here (maybe just for me, not sure how steeped you all are in this). If a library like ui-kit (or elements, which is a little more relevant) creates extra chunks to lazy import big libraries, are there reasonably easy ways to interop this in downstream consumers? E.g. if platform-internal, that uses webpack, imports ui-kit or elements that includes lazy load chunks, will things "just work"?

P0lip:
> creates extra chunks to lazy import big libraries

this chunk is a plain module (cjs or esm, but doesn't matter), so it doesn't really differ much from any other exported file.
Obviously, there is some webpack runtime crap added and all external deps are resolved, but that's about it.

>  E.g. if platform-internal, that uses webpack, imports ui-kit or elements that includes lazy load chunks, will things "just work"?

they should, because they import plain modules, and Webpack or any other robust bundler such as Rollup.js are able to determinate them at the build-time, and act accordingly (include them in the same chunk, create a separate chunk, etc.).
In other words - they do not differ from regular imports.
The story is by far different with monaco, for instance, where they use their own loader and workers they load cannot be statistically determined.


I tested it in platform-internal with yalced ui-kit. The thing here is that we do not retain `import` in the code, because we have `commonjs` target set, hm :/ 

@marbemac is there really any strong reason for us relying on commonjs in our own packages?
It's 2020 after all. :laughing:

P0lip:
`const CodeViewerComponent = React.lazy(() => Promise.resolve().then(() => require('./CodeViewer')));` here is how the output looks like. This will bailout any optimization, but there doesn't seem to be a way for TS to retain `import` call when you have `commonjs` module set.

marcelltoth:
> 
> 
> @marbemac is there really any strong reason for us relying on commonjs in our own packages?
> It's 2020 after all. 😆

I tried that once already. The problem is that we don't bundle our SSR server build with webpack in ninja, it is instead ran directly via `ts-node`. This is a problem because node - even v14 - does not support ES Modules the way webpack does. (I experimented with it a little back then, the thing is, node's new ESM support is still very finicky and requires a lot of attention when writing code for it, kind of defeating the purpose.) So unless we change that or get rid of SSR altogether, we need a commonjs build.

Now packing both a CJS **and** an ESM build is perfectly doable, tracked here: https://github.com/stoplightio/ui-kit/issues/165 I have experimented with various folder structures, etc, I believe I found the most convenient solution, let's talk if you'd like to make that improvement.

> I tested it in platform-internal with yalced ui-kit. The thing here is that we do not retain import in the code, because we have commonjs target set, hm :/

CommonJS just does not support that. It's not even ES2015 modules, only ES2020.

P0lip:
> This is a problem because node - even v14 - does not support ES Modules the way webpack does. (I experimented with it a little back then, the thing is, node's new ESM support is still very finicky and requires a lot of attention when writing code for it, kind of defeating the purpose.)

Node did ESM right IMHO - they addressed lots of issues, like implicit `index` resolving  (`import foo from './a/') or they require you to provide file extensions now, etc.
Webpack comes with its own module resolver, but we could tweak it, i.e. enforce extensions or modify aliases we use.
Last, but not least, enhanced-resolve supports plugins, soooo we could potentially write our own (if it doesn't exist yet) that would help us be more consistent.
I use native esm in all of my new private projects and it's great. It's mostly tooling that hasn't adjusted yet and the code you'd need to adapt.
CommonJS has its flaws as well, in particular the lack of support for `default`, that requires you to use interoping.

Either way, something to consider in future.

> I tested it in platform-internal with yalced ui-kit. The thing here is that we do not retain import in the code, because we have commonjs target set, hm :/

Yeah, I know it doesn't. I mean, they rightfully get rid of it, but I'd like to have it retained and there doesn't seem to be an option to do so.

P0lip:
@marcelltoth @wmhilton 
Soo, shall we stick to b183a20c44ad9c8899a50a8f6b4e47d72811c197 or revert b183a20c44ad9c8899a50a8f6b4e47d72811c197?
React.lazy makes no sense with CJS set as the target.
I'd really wish we used ESM (doesn't need to follow Node.js' ESM)  for all our Stoplight deps, but I believe this is a bit too much to ask for the time being.

wmhilton:
What happened to inlining the workers in the compiled output?

P0lip:
@wmhilton the worker is not a concern - everything was fine in this case. It's the `require` call that was retained and therefore ending up as showstopper for code-splitting

P0lip:
Decided to revert the revert. Having react.lazy will be beneficial once we use ESM (and we will eventually).
@wmhilton or @marcelltoth mind having a pass on it?

webpack.config.js
```diff
@@ -0,0 +1,40 @@
...
+    extensions: ['.tsx', '.ts', '.js', '.mjs'],
+  },
+
+  optimization: {
+    minimizer: [
+      new TerserPlugin({
+        terserOptions: {
+          ecma: 2020,
+          module: true,
+          safari10: true,
```
wmhilton, line 26: HahaHAHAHA I wondered why you brought in TerserPlugin lol.

wmhilton:
Do you have a draft PR pulling this into `platform-internal`?

wmhilton:
I like the code, but want to see that this worker crap actually works with Ninja and Desktop before I 👍

webpack.config.js
marbemac: FYI - we don't need to support safari 10 (if that is actually what this does).

wmhilton:
Tests failed

wmhilton:
This is one of the coolest things I've seen in our codebase in a long time. (I haven't looked at the actual code, just the functionality.) I _can_ get it to mess up sometimes though if I scroll to a random point in the document and then "scroll" up using the Page Up key. Check it out:

P0lip:
> I can get it to mess up sometimes though if I scrollbar to a random point in the document so the area above it is not highlighted and then "scroll" up using the Page Up key. Check it out:


did you resize the viewport beforehand or did anything else that could affect the functionality?
I'm struggling to reproduce it :/

wmhilton:
Nope I didn't resize the viewport. I can take another screen recording though if that helps.

wmhilton:
Hmm. Part of your problem might be you threw out my lovely storybook example. D:

Here ya go: an un-edited screen recording in all its glory. (Hopefully it's not too big to upload.) It takes a little luck - I think you have to scroll to some certain "fraction" of a page height or something.

[TODO: insert pic once I get the darn thing small enough to upload]

P0lip:
> Hmm. Part of your problem might be you threw out my lovely storybook example. D:

@wmhilton oh, did I?
I thought I pushed with `--force-with-lease`, so must have done it accidentally when merging or something.

wmhilton:
> Could you add the story again then?

Already did 😁

P0lip:
I think I can think of one case where this new approach might break, namely multi-line comments.
I don't think it's a big deal given the fact neither YAML nor JSON support them, and these are our main use cases.
Potentially multi-line strings might be some issue as well, but yet again - not a real issue in JSON, and rather not a huge deal in YAML world too.
Code samples should be totally fine, as they are small enough to be rendered in a single block.

PR #177, opened 2020-07-27
feat: add select option to toc component

mnaumanali94:


src/TableOfContents/index.tsx
```diff
@@ -198,7 +246,7 @@ export function TableOfContents<T extends TableOfContentsItem = TableOfContentsI
               {title}
             </Button>
           </div>
-          <div className="h-full flex justify-end">{comp}</div>
+          <div className="h-full flex flex-col justify-end">{comp}</div>
```
marbemac, line 249: Hmm - how did this work without `flex-col` before?

mnaumanali94: Hmm. Because there was just one component inside it? And this is for mobile.

PR #179, opened 2020-07-31
fix(code-viewer): reduce initial layout shift

P0lip:
This one has been caught by Marc.
before:
before
after:
after

### DIFF
--- a/src/styles/components/Code/_base.scss
+++ b/src/styles/components/Code/_base.scss
@@ -24,7 +24,6 @@
       margin-left: -$line-sidebar-width;
       padding-left: $line-sidebar-width;
       position: relative;
-      width: calc(100% - #{line-sidebar-width});
       display: block;
 
       &::before {
--- a/src/styles/components/Code/_base.scss
+++ b/src/styles/components/Code/_base.scss
@@ -11,15 +11,20 @@
   }
 
   &--line-numbers {
+    $line-sidebar-width: 45px;
+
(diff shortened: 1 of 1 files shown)


src/styles/components/Code/_base.scss
```diff
@@ -11,15 +11,20 @@
...
       margin-left: 35px !important;
     }
 
     .line-number {
-      padding-left: 45px;
+      margin-left: -$line-sidebar-width;
+      padding-left: $line-sidebar-width;
       position: relative;
-      width: calc(100% - 45px);
+      width: calc(100% - #{line-sidebar-width});
```
marcelltoth, line 27: This is a CSS syntax error. Resolves to `calc(100% - line-sidebar-width)`. You meant to do
```css
width: calc(100% - #{$line-sidebar-width});
```

marcelltoth: But. As you see it's not needed at all. The rows already expand as much as they want to, there is no point in forcing a 100% width, am I missing something?

marcelltoth: In case you did need it, it would be `100% + #{$line-sidebar-width}` anyway. Padding doesn't count because of `box-sizing: border-box`, but you need to make up for the negative margin.

P0lip: As discussed, I think it might have been needed by some browser, but we should be good now I suppose.

PR #180, opened 2020-07-31
feat(code-viewer): scroll overscan

P0lip:
This could have been done as a part of https://github.com/stoplightio/ui-kit/pull/179, but 179 is a no-brainer, while this one might potentially break something (it shouldn't, though).
It also improves support for initial scrollTop scenario.


wmhilton:
Doesn't seem to have broken anything. What's it supposed to do though? I can't tell.

wmhilton:
It doesn't fix the problem where stuff is briefly un-highlighted.

P0lip:
@wmhilton ready for another round.

wmhilton:
@P0lip Are we not using a ResizeObserver?

# Firefox



# Chrome

PR #181, opened 2020-08-02
fix(table-of-contents): apply maxWidth to selectFilter

mnaumanali94:
Table of Contents `selectFilter`:

Sets width for select and truncates large text instead of allowing it to wrap.

**Screenshots**

Currently in production:
Screenshot 2020-08-03 at 3 37 16 PM

After:
Screenshot 2020-08-03 at 3 38 35 PM

### DIFF
--- a/src/TableOfContents/index.tsx
+++ b/src/TableOfContents/index.tsx
@@ -214,6 +214,7 @@ export function TableOfContents<T extends TableOfContentsItem = TableOfContentsI
             className={cn(`m-${padding}`, { 'mr-0': hasFilter })}
           >
             <Button
+              className="w-48 truncate"
               text={getActiveItem()?.name ?? selectFilter.initialContent ?? 'Fetching...'}
               alignText="left"
               rightIcon="caret-down"
--- a/src/TableOfContents/index.tsx
+++ b/src/TableOfContents/index.tsx
@@ -214,7 +214,8 @@ export function TableOfContents<T extends TableOfContentsItem = TableOfContentsI
             className={cn(`m-${padding}`, { 'mr-0': hasFilter })}
           >
             <Button
-              className="w-48 truncate"
+              className="truncate"
+              style={{ maxWidth: '12rem' }}
               text={getActiveItem()?.name ?? selectFilter.initialContent ?? 'Fetching...'}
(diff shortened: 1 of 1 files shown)


src/TableOfContents/index.tsx
```diff
@@ -214,6 +214,7 @@ export function TableOfContents<T extends TableOfContentsItem = TableOfContentsI
             className={cn(`m-${padding}`, { 'mr-0': hasFilter })}
           >
             <Button
+              className="w-48 truncate"
```
marbemac, line 217: I suggest a max width, rather than width. If we don't have appropriate max width tailwind classes, can always just fall back to style if needed.

mnaumanali94: Yeah, we don't. Alright, style it is. This makes the UI less jittery though with changing lengths of org names.

marcelltoth:
What scenario does this improve?

Could you share a before / after screenshot pair?

mnaumanali94:
@marcelltoth 

Currently in production:
Screenshot 2020-08-03 at 3 37 16 PM

After:
Screenshot 2020-08-03 at 3 38 35 PM

src/TableOfContents/index.tsx
```diff
@@ -214,6 +214,8 @@ export function TableOfContents<T extends TableOfContentsItem = TableOfContentsI
             className={cn(`m-${padding}`, { 'mr-0': hasFilter })}
           >
             <Button
+              className="truncate"
+              style={{ maxWidth: '12rem' }}
```
marcelltoth, line 218: I'm not sure this second line here is applicable globally. This `12rem` is an arbitrary number that looks good in that single case in ninja, but may not look good for others. What about defining it as `flex-basis: 0; flex-grow: ?` on both components (this and the text filter).  You can maybe add a `minWidth` / `maxWidth` as well but I wouldn't make those the primary width attributes.

marcelltoth:
I'm closing this as it conflicts with @marbemac removing the feature in https://github.com/stoplightio/ui-kit/pull/183

If the problem persists on current master (after stoplightio/platform-internal#3881), you can open a PR against *platform-internal* that contains similar logic.

PR #182, opened 2020-08-03
feat(code-viewer): align line numbers to the right

P0lip:
Instead of pre-defined value, let's actually adjust the width according to the number of digits

Before:
image
image
image
image

After:
image
image
image
image
image
image

### DIFF
--- a/src/CodeViewer/components/BlockCodeViewer/BlockCodeViewer.tsx
+++ b/src/CodeViewer/components/BlockCodeViewer/BlockCodeViewer.tsx
@@ -20,6 +20,7 @@ const BlockCodeViewer: React.FC<IBlockCodeViewerProps> = ({ className, language,
   const [maxLines, setMaxLines] = React.useState<number | null>(null);
   const observerRef = React.useRef(new ObservableSet());
   const slicedBlocks = useSlicedBlocks(value, maxLines === null ? null : Math.max(0, maxLines - 1));
+  const linesNo = String(slicedBlocks !== null && maxLines !== null ? slicedBlocks.length * maxLines : 0).length;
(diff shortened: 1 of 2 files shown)


lottamus:
@johnrazmus this might fix your issue: https://github.com/stoplightio/platform-internal/issues/3266

src/CodeViewer/components/BlockCodeViewer/BlockCodeViewer.tsx
```diff
@@ -20,6 +20,7 @@ const BlockCodeViewer: React.FC<IBlockCodeViewerProps> = ({ className, language,
   const [maxLines, setMaxLines] = React.useState<number | null>(null);
   const observerRef = React.useRef(new ObservableSet());
   const slicedBlocks = useSlicedBlocks(value, maxLines === null ? null : Math.max(0, maxLines - 1));
+  const linesNo = String(slicedBlocks !== null && maxLines !== null ? slicedBlocks.length * maxLines : 0).length;
```
marcelltoth, line 23: This is not `linesNumber` but st like `lineNumberCharacterCount` or `lineNumberLength` or similar.

Naming is important, it's already non-trivial to understand what the different variables do.

src/styles/components/Code/_base.scss
```diff
@@ -11,28 +11,41 @@
   }
 
   &--line-numbers {
-    $line-sidebar-width: 45px;
-
-    padding-left: $line-sidebar-width;
+    $line-sidebar-width: 15px;
```
marcelltoth, line 14: This is now `line-number-char-width` or similar.

PR #185, opened 2020-08-06
feat(table-of-contents): rowComponentExtraProps

marcelltoth:
This PR adds a prop to the `TableOfContents` that will be passed down to the `rowComponent`.

The thing is type-safe (apart from that error which I couldn't find the cause of and *ts-ignore*d), it only lets you pass `rowComponentExtraProps` if the specified `rowComponent` requires them.

Check out the new stories for demonstration.

## Motivation
This allows us to avoid the very error prone action of creating components inside other components using `useCallback`.

As an example check out `ninja` on the `feat/custom-toc` branch: https://github.com/stoplightio/platform-internal/blob/feat/custom-toc/packages/ninja/src/screens/Workspace/Sidebar/Project.tsx#L323-L358

Currently there is no nice way to achieve that behavior, because the component relies on `itemPath`. But this is not how `rowComponent` is supposed to be used. (There are alternatives, such as passing `itemPath` via a private context, etc, but none of those are very elegant.)

With this feature we'll be able to write:
```typescript
type ExtraProps = {
  itemPath: (uri: string) => string;
}

const RowComponent: RowComponentType<TableOfContentsLinkWithId, ExtraProps> = ({item, extra: {itemPath}} => {
  // ...
});

// then in CustomProjectTableOfContents 
return (
      <TableOfContentsContainer
        workspaceUrl={`https://${state.router.hostname}`}
        projectSlug={projectSlug}
        branchSlug={branchSlug}
        urqlClient={urqlClient}
        onData={(tocTree: ITableOfContentsTree) => {
          if (tocTree?.items?.length) {
            const firstItem = tocTree.items.find(isItem);
            setFirstItem(firstItem);
          }
        }}
        rowComponent={RowComponent}
        rowComponentExtraProps={{itemPath}}
      />
    );
```

This is not only nicer but improves behavior and performance (no remount on `itemPath` change), and the `useCallback` pattern is very bad in that it's really easy to accidentally mess it up. Which will be very important once the library is released to the public.

### DIFF
--- a/src/TableOfContents/index.tsx
+++ b/src/TableOfContents/index.tsx
@@ -43,26 +43,26 @@ export type ITableOfContentsLink = TableOfContentsItem & {
   isExternalLink?: boolean;
 };
 
-export type RowComponentProps<T extends TableOfContentsItem> = {
+export type RowComponentProps<T extends TableOfContentsItem, E = {}> = {
   item: T;
   index: number;
   isExpanded: boolean;
   toggleExpanded: () => void;
-};
+} & ({} extends E ? {} : { extra: E });
 
-export type RowComponentType<T extends TableOfContentsItem> = React.ComponentType<RowComponentProps<T>>;
+export type RowComponentType<T extends TableOfContentsItem, E = {}> = React.ComponentType<RowComponentProps<T, E>>;
 
-export interface ITableOfContents<T extends TableOfContentsItem = TableOfContentsItem> {
+export type ITableOfContents<T extends TableOfContentsItem = TableOfContentsItem, E = {}> = {
   contents: T[];
-
-  /**
-   * Optionally customize how a row is rendered. Defaults to `DefaultRow`.
-   */
-  rowComponent?: RowComponentType<T>;
 
   // Padding that will be used for (default: 10)
   padding?: string;
   className?: string;
+
+  /**
+   * Optionally customize how a row is rendered. Defaults to `DefaultRow`.
+   */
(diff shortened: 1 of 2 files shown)


src/TableOfContents/index.tsx
```diff
@@ -42,27 +42,27 @@ export type ITableOfContentsLink = TableOfContentsItem & {
...
   isExpanded: boolean;
   toggleExpanded: () => void;
-};
+} & ({} extends E ? {} : { extra: E });
 
-export type RowComponentType<T extends TableOfContentsItem> = React.ComponentType<RowComponentProps<T>>;
+export type RowComponentType<T extends TableOfContentsItem, E = {}> = React.ComponentType<RowComponentProps<T, E>>;
 
-export interface ITableOfContents<T extends TableOfContentsItem = TableOfContentsItem> {
+export type ITableOfContents<T extends TableOfContentsItem = TableOfContentsItem, E = {}> = {
```
mallachari, line 54: That should probably be renamed to `TableOfContentsType` but that would be a breaking change so I guess better keep it for now.

marcelltoth: You're right. I'm not a great fan of the `I` prefix no matter if it's an interface or a type (there's barely any difference anymore) because IMO it brings no value and generally advised against in official TS sources (just google "typescript i prefix").

I'd really just name it `TableOfContentsProps`, that seems to be the convention in most libs, but as you guessed I didn't want to make a breaking change here.

PR #188, opened 2020-08-11
fix(code-viewer): get rid of trailing lines

P0lip:
Fixes https://github.com/stoplightio/platform-internal/pull/3901#issuecomment-671279905

I realllly hope this is the last PR :laughing:

### DIFF
--- a/src/CodeViewer/components/BlockCodeViewer/hooks/useSlicedBlocks.ts
+++ b/src/CodeViewer/components/BlockCodeViewer/hooks/useSlicedBlocks.ts
@@ -1,33 +1,37 @@
 import * as React from 'react';
 
-class SlicedBlocks extends Array<string> {
-  public length = 1;
-  public [0] = '';
-  public linesMap!: Map<number, number>;
+type SlicedBlock = {
+  id: string;
+  value: string;
+  lineCount: number;
+};
+
+function createSlicedBlock(): SlicedBlock {
+  return {
+    id: Math.random().toString(36),
+    value: '',
+    lineCount: 0,
+  };
 }
 
-Object.defineProperty(SlicedBlocks.prototype, 'linesMap', {
-  value: new Map<number, number>(),
-});
-
 export const useSlicedBlocks = (value: string, maxLines: number | null) => {
-  return React.useMemo<SlicedBlocks | null>(() => {
+  return React.useMemo<SlicedBlock[] | null>(() => {
     if (maxLines === null) {
       return null;
     }
 
-    const blocks = new SlicedBlocks();
+    const blocks: SlicedBlock[] = [createSlicedBlock()];
 
     for (let i = 0, n = 0; i < value.length; i++) {
       const char = value[i];
-      blocks[blocks.length - 1] += char;
+      blocks[blocks.length - 1].value += char;
 
       if (char === '\n') {
         n++;
 
         if (n % maxLines === 0 && i + 1 !== value.length) {
-          blocks.linesMap.set(blocks.length - 1, n);
-          blocks.push('');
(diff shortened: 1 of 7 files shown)


lottamus:
I yalced this into platform-internal and it's still doing a weird jump



And also not highlighting all of the lines



And for comparison, here's how the indentation looking in production

image

@P0lip @wmhilton how about we switch to an open source code highlighter like https://github.com/react-syntax-highlighter/react-syntax-highlighter

P0lip:
@lottamus 
the jump from left to right is on purpose.
We adjust the size of sidebar so that all numbers can fit.
Previously, we used to have a pre-defined width, but this is no longer the case.
Right now, the width is adjusted to the number of digits.

> And also not highlighting all of the lines

bahhhh, let me see why is this happening. :roll_eyes:

lottamus:
> We adjust the size of sidebar so that all numbers can fit.

I think we might be over adjusting in this case. I'm pasting in the platform-internal root package.json which only goes up to 3 digits. Maybe start the adjustment if we're at 5+ digits. Or even larger, I'd rather our default styling handles a majority of the use-cases so most users never see the jumping (unless they're looking at something juuuuuge)

P0lip:
Speaking of code highlighting issue - I know why it's happening.
It's because we currently listen to window and root scroll events, but in case of Studio we have that fancy custom scrollbar.

> I think we might be over adjusting in this case. I'm pasting in the platform-internal root package.json which only goes up to 3 digits. Maybe start the adjustment if we're at 5+ digits. Or even larger, I'd rather our default styling handles a majority of the use-cases so most users never see the jumping (unless they're looking at something juuuuuge)

Yeah, that makes sense. Will do.

wmhilton:
> @P0lip @wmhilton how about we switch to an open source code highlighter like https://github.com/react-syntax-highlighter/react-syntax-highlighter

@lottamus Well, `react-syntax-highlighter` uses the exact same library (`refractor`) that we do, so I doubt it would be able to handle these large files. I can throw together a POC to compare though.

src/CodeViewer/components/BlockCodeViewer/hooks/useSlicedBlocks.ts
```diff
@@ -1,22 +1,37 @@
...
-  return React.useMemo<string[] | null>(() => {
+  return React.useMemo<SlicedBlock[] | null>(() => {
     if (maxLines === null) {
       return null;
     }
 
-    const blocks: string[] = [''];
+    const blocks: SlicedBlock[] = [createSlicedBlock()];
 
     for (let i = 0, n = 0; i < value.length; i++) {
```
wmhilton, line 25: image
image

P0lip:
@wmhilton do yo uhave the latest elements installed?
I fixed that issue in elements. It's not codeviewer's fault. It's TryIt component using `safeStringify` that returns undefined when you pass undefined.

wmhilton:
Ahhh. So I gotta upgrade both at once huh? :eyeroll: OOOOOk lemme try that.

P0lip:
For you reference
https://github.com/stoplightio/elements/pull/514
https://github.com/stoplightio/platform-internal/pull/3906

wmhilton:
> @lottamus Well, `react-syntax-highlighter` uses the exact same library (`refractor`) that we do, so I doubt it would be able to handle these large files. I can throw together a POC to compare though.

@lottamus Here's you answer:

Firefox:
image


Chrome:



At the time I post this, my Chrome tab still hasn't recovered. My fans are spinning like crazy. Google Chrome Helper (Renderer) is using 100% CPU.

PR #189, opened 2020-08-18
fix(table-of-contents): typing issue

marcelltoth:
Fixes https://github.com/stoplightio/platform-internal/pull/3901#issuecomment-675145014

### DIFF
--- a/src/TableOfContents/index.tsx
+++ b/src/TableOfContents/index.tsx
@@ -2,6 +2,7 @@ import { Button, Drawer } from '@blueprintjs/core';
 import cn from 'classnames';
 import { flatMap, range } from 'lodash';
 import * as React from 'react';
+import { PropsWithChildren, ReactElement } from 'react';
 
 import { FAIcon, FAIconProp } from '../FAIcon';
 import { ScrollContainer } from '../ScrollContainer';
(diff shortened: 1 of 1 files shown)


src/TableOfContents/index.tsx
```diff
@@ -49,7 +50,10 @@ export type RowComponentProps<T extends TableOfContentsItem, E = {}> = {
   toggleExpanded: () => void;
 } & ({} extends E ? {} : { extra: E });
 
-export type RowComponentType<T extends TableOfContentsItem, E = {}> = React.ComponentType<RowComponentProps<T, E>>;
+export type RowComponentType<T extends TableOfContentsItem, E = {}> = (
+  props: PropsWithChildren<RowComponentProps<T, E>>,
+  context?: any,
+) => ReactElement<any, any> | null;
```
marcelltoth, line 56: This is a subset of the original `React.ComponentType`.

Unforunately `ComponentType` / `FunctionComponent` do not play nice with generics because of `propTypes`. This is the call signature extracted from `FunctionComponent` which is enough for our purposes yet enables generics.

mallachari: Maybe a comment above so no one need to wonder why it is done this way?

PR #190, opened 2020-08-19
fix(code-viewer): remove Suspense

P0lip:
See https://github.com/stoplightio/platform-internal/pull/4035#discussion_r473085845 for the reasoning behind the change


wmhilton:
Keep the file named

> I would keep the file named `src/CodeViewer/components/BlockCodeViewer/BlockCodeViewer.tsx` and just update `index.ts` to export it. Might make the diff readable.

See?

before: 
image

after: 
image

PR #192, opened 2020-08-20
fix(code-viewer): use useEffect instead of useLayoutEffect

P0lip:
`useLayoutEffect` is not supported by SSR.
Swapping `useLayoutEffect` with `useEffect` improves the rendering of CodeViewer on sites that use SSR.
The regular usage does not seem to be notably affected by this change.

### DIFF
--- a/src/CodeViewer/components/BlockCodeViewer/BlockCodeViewer.tsx
+++ b/src/CodeViewer/components/BlockCodeViewer/BlockCodeViewer.tsx
@@ -29,7 +29,7 @@ const BlockCodeViewer: React.FC<IBlockCodeViewerProps> = ({ className, language,
     slicedBlocks !== null && maxLines !== null ? slicedBlocks.length * maxLines : 0,
   ).length;
 
-  React.useLayoutEffect(() => {
+  React.useEffect(() => {
     if (nodeRef.current !== null) {
       setMaxLines(calculateMaxLines(window.innerHeight)); // we have to use window here, as element may not ave any height at this time
     }


wmhilton:
What do we need to do to enable SSR for CodeViewer in platform? I tried yalcing it but as you can see, no bueno.




done using:
image

wmhilton:
To clarify... I would expect the text inside the CodeViewer to be present before hydration, and only the syntax highlighting applied client side in the `useEffect`.

wmhilton:
Aha. Figured it out, sort of. If I change https://github.com/stoplightio/ui-kit/pull/192/files#diff-53d277950c0626d41529ff6a0798bc79R24 from `null` to say, `100`, then I see the desired result.



@P0lip Does `maxLines` even matter? It just determines the batch size of how many lines are highlighted at a time right? Can we just use a constant like... 10? Or 1?

wmhilton:
@P0lip Don't be mad, but I simplified your code a bit. 😄 It works well with the "lots of text" example in Storybook as well as yalced into platform.

wmhilton (dismissed):
:trollface: I APPROVE MY CODE

P0lip:
Hold on

P0lip:
> What do we need to do to enable SSR for CodeViewer in platform? I tried yalcing it but as you can see, no bueno.

Hmm, it works differently in my case. :man_shrugging: 
Did you set resolution? I suppose you did, but just making sure.

In any case, I'd really refrain from changing maxLines.
I added it for a reason. 
It's here to make sure we only highlight the relevant part and not N lines.
If your screen is capable of rendering 400 lines, the maxLines would equal ~401.
The change you made may cause the scroll flickering you @wmhilton hated.
Right now, we highlight the visible block as well as its siblings,see
https://github.com/stoplightio/ui-kit/blob/9b29cfda5dcdca20591af0d91d9b8bb66d82aaf2/src/CodeViewer/components/BlockCodeViewer/BlockCodeViewer.tsx#L38

wmhilton:
> It's here to make sure we only highlight the relevant part and not N lines.

Yes but we highlight all visible blocks + siblings. So there's really no relationship to the viewport.

P0lip:
yeah, but this is only because we use IntersectionObserver.
In any case, I don't really mind. If it works okay, let's ship it.

wmhilton:
> this is only because we use IntersectionObserver.

Exactly. Using IntersectionObserver basically deprecates all the old logic. It's just much better. 🥳 

Give it a whirl, I honestly think it's worth it because we can remove ResizeObserver completely and reduce a lot of complexity.

P0lip:
It's okay. Less code is usually better.

PR #193, opened 2020-08-24
feat(code-viewer): animated transition from plain to highlighted

marcelltoth:
Followup to #192 and stoplightio/platform-internal#4086

## Summary

Animates the transition from plain text view to highlighted view.

## Details
When first rendering the CodeViewer it does not perform syntax highlighting.

To optimize performance highlighting happens after rendering in a `useEffect` hook. This causes a flicker when the coloring is finally applied to the text. This is especially noticable when the initial render is performed server-side.

This fix smoothens out the process using a short (0.3s) animation to introduce the color.

### Why not inside CodeEditor?

The library CodeEditor is using is probably missing some keys, which makes every token `div` remounted on change (which happens quite often in an editor). This causes repeated animations.

### Extra

Fixed that ugly `as any` cast.

## Demo

Check out the codeviewer storybooks.


P0lip (approved):
Seems harmless.
Tested it and everything works correctly.

PR #194, opened 2020-10-01
feat: bump fontawesome types to latest version

marcelltoth:
To remove version conflict and resolution in https://github.com/stoplightio/elements/pull/607

Only types changed.


mmiask (approved):
Looks good!

PR #195, opened 2020-11-16
feat: Display endpoint method and icons for schema and API

mallachari:
Needed by https://github.com/stoplightio/elements/issues/639

Changes:
- fixes collapsing all descendant children nodes when toggling parent (nested nodes remained open - I've added a simple story for that)
- adds toc item icons positioning (we need to have them on both sides)
- allows using text labels instead of icons (i.e. for oas operation labels)
- splits group items containing `to` property so they can have both actions handled

~~Note: note sure if we should use paid FA icons here (as in original issue). IIRC elements were supposed to use free version which leaves platform without a choice.~~

Peek 2020-11-16 04-44

### DIFF
--- a/.storybook/preview-head.html
+++ b/.storybook/preview-head.html
@@ -1,0 +1,3 @@
+<!-- .storybook/preview-head.html -->
+
+<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/5.15.1/css/all.min.css">
--- a/src/TableOfContents/index.tsx
+++ b/src/TableOfContents/index.tsx
@@ -110,16 +110,14 @@ function TableOfContentsInner<T extends TableOfContentsItem = TableOfContentsIte
   // an array of functions. Invoking the N-th function toggles the expanded flag on the N-th content item
   const toggleExpandedFunctions = React.useMemo(() => {
     return range(contents.length).map(i => () => {
-      let childrenToCollapse = {};
-      if (expanded[i]) {
-        const item = contents[i];
-        const children = findDescendantIndices(item.depth ?? 0, i, contents.slice(i + 1));
-        childrenToCollapse = children.reduce((obj, index) => {
-          obj[index] = false;
-          return obj;
-        }, {});
-      }
       setExpanded(current => {
+        let childrenToCollapse = {};
+        if (current[i]) {
+          const item = contents[i];
+          const children = findDescendantIndices(item.depth ?? 0, i, contents.slice(i + 1));
+          childrenToCollapse = Object.fromEntries(children.map(i => [i, false]));
+        }
+
         return {
           ...current,
           [i]: !current[i],
@@ -128,7 +126,7 @@ function TableOfContentsInner<T extends TableOfContentsItem = TableOfContentsIte
       });
     });
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [contents, contents.length, expanded]);
+  }, [contents, contents.length]);
 
   // expand ancestors of active items by default
   React.useEffect(() => {
@@ -347,7 +345,7 @@ function DefaultRowImpl<T extends TableOfContentsItem>({ item, isExpanded, toggl
           {actionElem}
           {item.iconPosition === 'right' && iconElem}
           {isGroup && (
-            <div onClick={() => isGroupItem && toggleExpanded()} className="px-2">
+            <div onClick={isGroupItem ? toggleExpanded : undefined} className="px-2">
               <FAIcon className="TableOfContentsItem__icon" icon={isExpanded ? 'chevron-down' : 'chevron-right'} />
             </div>
           )}
--- a/src/TableOfContents/index.tsx
+++ b/src/TableOfContents/index.tsx
@@ -109,13 +109,26 @@ function TableOfContentsInner<T extends TableOfContentsItem = TableOfContentsIte
 
   // an array of functions. Invoking the N-th function toggles the expanded flag on the N-th content item
   const toggleExpandedFunctions = React.useMemo(() => {
-    return range(contents.length).map(i => () =>
-      setExpanded(current => ({
-        ...current,
-        [i]: !current[i],
-      })),
-    );
-  }, [contents.length]);
+    return range(contents.length).map(i => () => {
+      let childrenToCollapse = {};
+      if (expanded[i]) {
+        const item = contents[i];
+        const children = findDescendantIndices(item.depth ?? 0, i, contents.slice(i + 1));
+        childrenToCollapse = children.reduce((obj, index) => {
+          obj[index] = false;
+          return obj;
+        }, {});
+      }
+      setExpanded(current => {
+        return {
+          ...current,
+          [i]: !current[i],
+          ...childrenToCollapse,
+        };
+      });
+    });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [contents, contents.length, expanded]);
 
   // expand ancestors of active items by default
   React.useEffect(() => {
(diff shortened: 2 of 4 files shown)


.storybook/preview-head.html
```diff
@@ -0,0 +1,3 @@
+<!-- .storybook/preview-head.html -->
+
+<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/5.15.1/css/all.min.css">
```
mpodlasin, line 3: This is interesting to me.

Does it mean in production we are also loading those styles from cdn?

When I used `font-awesome` icons last time I remember, React version was including the styles by itself (or at the very least they were coming from `node_modules` package, not cdn).

.storybook/preview-head.html
```diff
@@ -0,0 +1,3 @@
+<!-- .storybook/preview-head.html -->
```
mpodlasin, line 1: Interestingly, when running storybook, I get an error:

Screenshot from 2020-11-16 15-38-31

If this is not caused by your PR or it only appears on my computer, you can ignore this comment.

mpodlasin (changes requested):

.storybook/preview-head.html
mallachari: In production we don't need to do that this way, but icons in storybook here and in elements don't work (at least for me) so I added this which helped. Though I didn't put much attention to that, it could probably be done differently, I just wanted to be able to test the pr

mpodlasin:
I don't known why, but in my storybook, in `nested` story, clicking on a topmost button does nothing. I have to click on the arrow itself to make it work.

mallachari:
> I don't know why, but in my storybook, in `nested` story, clicking on a topmost button does nothing. I have to click on the arrow icon specifically to make it work.
> 
> Interestingly all the other buttons, apart from topmost one, work as expected. 
> 
> And it happens only in `nested` story, other stories work fine.

That is the feature :D 
It's group item with link so clicking it unlike others navigates to node and only arrow expands it.

Check out the requirements for Project ToC https://github.com/stoplightio/elements/issues/638

It says:
> When generating the TableOfContents for an API, add the API as a group with tags, endpoints, and schemas nested inside of it
Clicking the API should navigate to API overview, and not toggle expand/collapse
Clicking "caret" icon for API should toggle the group expanded or collapsed

src/TableOfContents/index.tsx
```diff
@@ -107,13 +109,26 @@ function TableOfContentsInner<T extends TableOfContentsItem = TableOfContentsIte
...
+          return obj;
+        }, {});
+      }
+      setExpanded(current => {
+        return {
+          ...current,
+          [i]: !current[i],
+          ...childrenToCollapse,
+        };
+      });
```
marcelltoth, line 128: There was some reason I moved some of the logic to inside the callback function. Likely it's safer there, but in any case, I'd rather have it at a single place. What do you think of
```tsx
      setExpanded(current => {
        let childrenToCollapse = {};
        if (current[i]) {
          const item = contents[i];
          const children = findDescendantIndices(item.depth ?? 0, i, contents.slice(i + 1));
          childrenToCollapse = children.reduce((obj, index) => {
            obj[index] = false;
            return obj;
          }, {});
        }

        return {
          ...current,
          [i]: !current[i],
          ...childrenToCollapse,
        };
      });
```

And while you're there you could consider
```tsx
childrenToCollapse = Object.fromEntries(children.map(i => [i, false]));
```
which is the same as your `reduce`.

src/TableOfContents/index.tsx
```diff
@@ -107,13 +109,26 @@ function TableOfContentsInner<T extends TableOfContentsItem = TableOfContentsIte
...
+    });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [contents, contents.length, expanded]);
```
marcelltoth, line 131: Oh this is one of the reasons why.
If you make the above change you'll only depend on `contents` and can get rid of the eslint inhibit.

src/TableOfContents/index.tsx
```diff
@@ -19,6 +19,8 @@ export type TableOfContentsItem = {
   icon?: FAIconProp;
   activeIcon?: FAIconProp;
   iconColor?: string;
+  iconPosition?: 'left' | 'right';
+  textIcon?: string;
```
marcelltoth, line 23: It's starting to feel like our abstraction is broken: there's way too many `icon` props allowing for not too many use-cases.
Not something to block this PR, but next time we touch ToC this will likely need some cleanup.

src/TableOfContents/index.tsx
```diff
@@ -313,11 +345,11 @@ function DefaultRowImpl<T extends TableOfContentsItem>({ item, isExpanded, toggl
           {item.meta && <span className="text-sm text-left text-gray font-medium">{item.meta}</span>}
           {loadingElem}
           {actionElem}
+          {item.iconPosition === 'right' && iconElem}
           {isGroup && (
-            <FAIcon
-              className="TableOfContentsItem__icon"
-              icon={['far', isExpanded ? 'chevron-down' : 'chevron-right']}
-            />
+            <div onClick={() => isGroupItem && toggleExpanded()} className="px-2">
```
marcelltoth, line 350: IMO a nicer way to express this would be `onClick={isGroupItem ? toggleExpanded : undefined}`. A tiny bit more efficient as well but mainly more expressive:

*If it's a group item then we handle the click like this. Otherwise we **don't** handle it at all.*

marcelltoth (approved):
>> I don't know why, but in my storybook, in `nested` story, clicking on a topmost button does nothing. I have to click on the arrow icon specifically to make it work.
>> 
>> Interestingly all the other buttons, apart from topmost one, work as expected. 
>> 
>> And it happens only in `nested` story, other stories work fine.
>
>That is the feature :D 
>It's group item with link so clicking it unlike others navigates to node and only arrow expands it.

TBH I completely agree with @mpodlasin 's opinion on this. It's horrible user experience. The same user action "clicking an item" that you are used to suddenly "just doesn't work" on some items, without having a clear indicator on *why* that is the case. Especially true for the API Overview page which is the default page, so trying to expand that node will be a **complete no-op** for the user as they are already on that page.

Having said that, this is a design issue. The PR delivers what was asked for and is very important to finish the rest of the issues so gets the ✔️ from me. But cc @philsturgeon we'll likely need a followup adjustment here.

 I left some comments but none of them are blockers. 🚢 it!

.storybook/preview-head.html
marcelltoth: I know the FA situation is still weird. 

Currently `ui-kit` doesn't contain FontAwesome itself but documents it as a dependency to be installed by its users. What @mallachari did here was what we usually do in storybooks to satisfy such deps, in SaaS production we install from `node_modules` as you said.

One reason you don't want to be installing FA in a library as a regular dep is that many sites already use it (mostly using some CDN), and bundling it would result in duplication and possible version conflicts. We can discuss a better FA strategy on Slack if you have ideas, this is of course by no means perfect.

src/TableOfContents/index.tsx
mallachari: Yup, that would be good, I've changed it. 
`fromEntries` indeed might be better - more readable too.

src/TableOfContents/index.tsx
mallachari: I think we still would need that `contents.length`, just `contents` would be enough to react on length change, right?

src/TableOfContents/index.tsx
mallachari: Good point. Done

mallachari:
I also had these thoughts about that but that's some kind of trade off for not having `Overview` item. In actual ToC it feels better as clicking it has an action of switching node.

PR #197, opened 2020-11-17
feat: toc start expanded

mallachari:
Needed by https://github.com/stoplightio/elements/pull/722

Adds support for initially expanded ToC items.

### DIFF
--- a/src/TableOfContents/index.tsx
+++ b/src/TableOfContents/index.tsx
@@ -106,7 +106,10 @@ function TableOfContentsInner<T extends TableOfContentsItem = TableOfContentsIte
   rowComponent: RowComponent = DefaultRow,
   rowComponentExtraProps,
 }: Pick<ITableOfContents<T, E>, 'className' | 'contents' | 'rowComponent' | 'rowComponentExtraProps'>) {
-  const [expanded, setExpanded] = React.useState({});
+  const [expanded, setExpanded] = React.useState(() => {
+    const itemsToExpand = contents.filter(item => item.startExpanded);
+    return Object.fromEntries(itemsToExpand.map(item => [contents.indexOf(item), true]));
+  });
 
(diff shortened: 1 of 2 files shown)


src/TableOfContents/index.tsx
```diff
@@ -128,6 +129,21 @@ function TableOfContentsInner<T extends TableOfContentsItem = TableOfContentsIte
...
+      findDescendantIndices(item.depth ?? 0, contents.indexOf(item), contents.slice(contents.indexOf(item) + 1)),
+    );
+
+    setExpanded(current => ({
+      ...current,
+      ...Object.fromEntries(itemsToExpand.map(index => [index, true])),
+      ...Object.fromEntries(childrenToExpand.map(index => [index, true])),
+    }));
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
marcelltoth, line 145: I don't buy why a prop called `startExpanded` would make children expanded.

Also is that required to adhere to the spec? It only wants you to expand individual APIs, not groups within, so that should be enough.

marcelltoth: And instead of hacking a `useEffect`, you could move the logic to line 108:
```ts
  const [expanded, setExpanded] = React.useState({});
```
and replace the empty initializer

marcelltoth (changes requested):

src/TableOfContents/index.tsx
mallachari: I understood `Apis should start expanded` as fully expanding its content with groups, but it might be just first level depth.
If we don't need to expand children then it will be simpler and without useEffect

marcelltoth: I don't think you need it all the way, that can get very messy quick

mallachari: Updated to be less greedy when expanding

marcelltoth (approved):
👋

